PR #14, opened 2018-06-22
Skip .data sections in .text. Fix for OpenSSL asm code

xpahos:
OpenSSL has code like this:

```
.globl	Camellia_EncryptBlock_Rounds
--
.type	Camellia_EncryptBlock_Rounds,@function
.align	16
.Lenc_rounds:
Camellia_EncryptBlock_Rounds:
pushq	%rbx
pushq	%rbp
pushq	%r13
pushq	%r14
pushq	%r15
.Lenc_prologue:
 
 
movq	%rcx,%r13
movq	%rdx,%r14
 
shll	$6,%edi
leaq	.LCamellia_SBOX(%rip),%rbp
leaq	(%r14,%rdi,1),%r15
```

Where LCamellia_SBOX is data:

```
.LCamellia_SBOX:
--
.long	0x70707000,0x70700070
.long	0x82828200,0x2c2c002c
.long	0x2c2c2c00,0xb3b300b3
.long	0xececec00,0xc0c000c0
```

This patch prevents adding this symbols to InterproceduralReferences set.

### DIFF
--- a/src/BinaryContext.h
+++ b/src/BinaryContext.h
@@ -207,6 +207,10 @@ public:
   /// when a function has more than a single entry point.
   std::set<uint64_t> InterproceduralReferences;
 
+  /// Set of pointers to .data segments found in .text section
+  /// This is workaround for plain assembly segments(e.g OpenSSL crypto functions)
+  std::set<BinaryData *> DataInTextPointers;
+
   std::unique_ptr<MCContext> Ctx;
 
   std::unique_ptr<DWARFContext> DwCtx;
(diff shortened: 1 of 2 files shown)


maksfb:
Thanks for the proposed fix! Did it work for you in non-relocation mode? In general, data in code is truly bad for performance. We have to teach BOLT to automatically move it out. Meanwhile a good workaround is an asm modification whenever possible. It will also help people who don’t use BOLT.

xpahos:
@maksfb I think I should close this pull request because on other two cases this workaround doesn't work.  Binary linked with OpenSSL has K_XX_XX symbols, which LLVM treats as functions, but actually, it's data.

> In general, data in code is truly bad for performance. 
I'm not so good at cryptography. I don't understand why OpenSSL uses data in this way. 

I will continue investigations. Hope I will find workaround better, than skip OpenSSL function names by comparing names

PR #18, opened 2018-06-26
In some cases DB could be nullptr

xpahos:
When processing binary with -debug mode in some cases, BD could be nullptr. It will be better to fail later on assert than here with segfault.


rafaelauler:
LGTM

Issue #15, opened 2018-06-22
Zero-sized symbol on end of section


shenhanc78:
Hi, we got an internal error for a symbol sitting right on the edge of section end.
```
  [13] xxx_malloc        PROGBITS        00000000238aa800 238aa800 001c7a 00  AX  0   0 64
  [14] malloc_hook       PROGBITS        00000000238ac480 238ac480 0006c2 00  AX  0   0 16
```
And we have a symbol with size 0 sitting exactly on the end of the section.
  xxx at 0x238ac47a = (00000000238aa800 + 001c7a)
So BinaryContext::getSectionFromAddress fails to find a home for the symbol.

The fix is straightforward, does it make sense to you to include zero-sized symbol on edge of section end?

PR #19, opened 2018-06-26
Refactor "getSectionForAddress" and make it accept symbols on end of sections

shenhanc78:
Refactor "getSectionForAddress" (extract common code into a separate function) and make it work for symbols sitting on end of sections.

The CL is to fix senerios that for some huge binaries there are symbols sitting on the end edge of a section, like below:
  [13] xxx_malloc        PROGBITS        00000000238aa800 238aa800 001c7a 00  AX  0   0 64
  [14] malloc_hook       PROGBITS        00000000238ac480 238ac480 0006c2 00  AX  0   0 16
xxx_malloc is at 0x238ac47a = (00000000238aa800 + 001c7a)
So BinaryContext::getSectionFromAddress fails to find a home for it.

PR: https://github.com/facebookincubator/BOLT/issues/15


rafaelauler:
Thanks Han Shen for the proposed fix. If you can think of any way to add the problem you solved as a testcase to our LIT infrastructure (see one of my last commits), that would be awesome, but that's OK to leave as is. We are still discussing a way of organizing the open-source tests.

PR #45, opened 2018-11-26
Fix URL in code of conduct

davidmalcolm:
https://code.facebook.com/pages/876921332402685/open-source-code-of-conduct doesn't work; it gives a "The content you’re looking for is not available at this URL."

Searching on code.fb.com led me to the URL in this pull request, which I'm assuming is the correct one for Facebook's open source projects.


maksfb:
The link on facebook.com was broken at some point but works now.

Issue #53, opened 2019-04-11
[C++] Use of tcmalloc leads to assertion failure


PierreRamoin-1A:
Hi,
I have a binary which use tcmalloc (from gperftools). tcmalloc creates some new sections for himself (google_malloc and malloc_hook) with AX flags.

When I apply BOLT on this binary, it leads to an assertion when meets the __stop_google_malloc symbol on
```
llvm::bolt::RewriteInstance::discoverFileObjects(): Assertion `Section && "section for functions must be registered."' failed.
```

If I force tcmalloc not to create this section, I don't have the assertion anymore.

I assume this have some links with #6, but as the issue moved a lot from the original subject, I preferred open a new one. 

Here is an extract of a readelf -e of libtcmalloc.a around the concerned sections:

```
  [...]
  [481] .gnu.lto_.symtab. PROGBITS         0000000000000000  000560d6
       0000000000000d72  0000000000000000   E       0     0     1
  [482] .gnu.lto_.opts    PROGBITS         0000000000000000  00056e48
       00000000000000bd  0000000000000000   E       0     0     1
  [483] .rodata           PROGBITS         0000000000000000  00056f20
       00000000000008a9  0000000000000000   A       0     0     32
  [484] malloc_hook       PROGBITS         0000000000000000  000577ca
       0000000000000211  0000000000000000  AX       0     0     2
  [485] .relamalloc_hook  RELA             0000000000000000  000a1010
       0000000000000228  0000000000000018   I      802   484     8
  [486] .gcc_except_table PROGBITS         0000000000000000  000579db
       000000000000003b  0000000000000000   A       0     0     1
  [487] .text._ZN4base6su PROGBITS         0000000000000000  00057a16
       0000000000000024  0000000000000000 AXG       0     0     1
  [488] .text._ZN4base6su PROGBITS         0000000000000000  00057a3a
       000000000000001c  0000000000000000 AXG       0     0     1
  [489] .text._ZN4base6su PROGBITS         0000000000000000  00057a56
       0000000000000023  0000000000000000 AXG       0     0     1
  [490] .rela.text._ZN4ba RELA             0000000000000000  000a1238
       0000000000000018  0000000000000018  IG      802   489     8
  [491] .text._ZN4base6su PROGBITS         0000000000000000  00057a79
       000000000000002f  0000000000000000 AXG       0     0     1
  [492] .rela.text._ZN4ba RELA             0000000000000000  000a1250
[...]

PR #55, opened 2019-05-10
Fix custom sections (#53)

PierreRamoin-1A:
Summary:
not process functions in custom sections,
and fix __start_/__stop_ section markers section process


src/RewriteInstance.cpp
```diff
@@ -2955,7 +2986,7 @@ void RewriteInstance::emitSections() {
 void RewriteInstance::emitFunctions(MCStreamer *Streamer) {
   auto emit = [&](const std::vector<BinaryFunction *> &Functions) {
     for (auto *Function : Functions) {
-      if (!BC->HasRelocations &&
+      if ((!BC->HasRelocations || Function->getSection().isProtected()) &&
```
maksfb: If there are references from functions in protected sections to functions in regular sections, how are these references going to get updated?

PierreRamoin-1A: I'm working on a fix for that but I have a question around it.
On this fix, I let BOLT processing the relocations on protected sections but I prevent moving the section itself, and all the functions inside of it.

But I still have in issue. BOLT is generating an other section (```.org.boltcustom```) to carry my ```__start_custom``` and my ```__stop_custom``` symbol.
Beside not  to be the reaction I want, this section is a perfect no-sense, as this section begin at ```__start_custom``` and have a size of 0. On the other hand, even if it had the "good" size, it would have been a non-sense either, as that means ```.org.boltcustom``` and ```custom``` would have been overlaid.

I tried a lot of things but I still have this section, either as ```.org.boltcustom``` or as a duplicate of ```custom```. My question is, do you know or can you give me an hint of how prevent this section creation?

maksfb: One idea is to handle immovable code sections as data. This was, the will be no need to even build `BinaryFunction`s for the contents. Of course the problem of updating the references still remains, and we have to rely on relocation info. The major exception from regular data sections is that there will be PC-relative relocations that need a special type of handling.

Romain-Geissler-1A: Hi @maksfb,

I am helping a bit Pierre on trying to make Bolt work on a binary linked against TcMalloc. My knowledge of Bolt is far below yours and the one of Pierre ;)

We don't understand fully what you mean. From what I get, you suggest us to handle the custom section as only raw data, but not functions, ok that point is fine. However as this section shall be translated by some offset in the final binary memory layout, we have to update:
 - any code outside of custom that calls any symbol defined in custom
 - any code inside custom that calls any symbol outside custom.

So we still need:
 - to list all symbols of custom and update the "normal" code referencing them
 - to update the code of custom itself (so parse it as functions and update the corresponding assembly).

Am I right ? So this doesn't really go in the direction of considering it fully as data.

From what Pierre told me, today everything works (ie "custom" memory layout is correctly kept, still functions inside it are modified accordinly when they reference foreign symbols), just the only problem is about this "__stop_" marker which is being wrongly associated to the following section/symbol, and thus relocated elsewhere. Do you have a simple advice for us on how exactly we should tackle this corner case in Bolt. Or any other clue we should try to dig in ? We aren't ELD experts as we would like to, and I know by experience explaining all these things takes a lot of time, but we will take all the clues you might give us to help us going in the right direction ;)

Cheers,
Romain

maksfb: Hi Romain and Pierre,

It's hard for me to tell exactly what is going on without looking at your other patch.

`.bolt.orgcustom` should not get created unless we are relocating `custom` section. With the current patch I cannot reproduce this problem, i.e. `custom` section stays in place.

You are correct listing items that need to be implemented. If you have almost everything working without it, then it might be a simpler route to take.

The `__stop` marker needs a special handling in `updateSymbolTable()`. If you can recognize it and skip its update then it should be enough.

Hope this helps. Again, it would be easier for me to assist you if I could reproduce the issue with the other patch you are talking about.

PR #96, opened 2020-08-12
Added option for specifying temporary files directory in perf2bolt tool

aagontuk:
By default, perf2bolt tool writes temporary files in /tmp.
But situation can arise when a different directory may have needed.
For example, there is not enough space in /tmp.
 `-tmpdir` option is added to do just that.

Example:
$ perf2bolt <executable> -p perf.data -tmpdir=$HOME/tmp/ -o perf.fdata

### DIFF
--- a/src/DataAggregator.cpp
+++ b/src/DataAggregator.cpp
@@ -144,6 +144,11 @@ WriteAutoFDOData("autofdo",
   cl::ZeroOrMore,
   cl::cat(AggregatorCategory));
 
+static cl::opt<std::string>
+TmpDirectory("tmpdir",
+  cl::desc("specify a directory to store temporary files."),
+  cl::cat(AggregatorCategory));
+
 }
 
 namespace {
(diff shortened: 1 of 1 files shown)


src/DataAggregator.cpp
```diff
@@ -264,22 +269,46 @@ void DataAggregator::launchPerfProcess(StringRef Name, PerfProcessInfo &PPI,
...
 
-  if (auto Errc = sys::fs::createTemporaryFile("perf.script", "out",
-                                               PPI.StdoutPath)) {
-    errs() << "PERF2BOLT: failed to create temporary file "
-           << PPI.StdoutPath << " with error " << Errc.message()
-           << "\n";
-    exit(1);
-  }
-  TempFiles.push_back(PPI.StdoutPath.data());
+  if(opts::TmpDirectory.empty()) {
```
rafaelauler, line 272: Ideally, use clang format (only on the lines that you are changing) for correct indentation.
`if (opts::TmpDirectory.empty())`

src/DataAggregator.cpp
```diff
@@ -264,22 +269,46 @@ void DataAggregator::launchPerfProcess(StringRef Name, PerfProcessInfo &PPI,
...
+
+    std::string ErrFileName(opts::TmpDirectory + "/perf.script-%%%%%%.err");
+    if (auto Errc = sys::fs::createUniqueFile(ErrFileName,
+                                                 PPI.StderrPath,
+                                                 sys::fs::owner_read | sys::fs::owner_write)) {
+      errs() << "PERF2BOLT: failed to create temporary file "
+             << PPI.StderrPath << " with error " << Errc.message() << "\n";
+      exit(1);
+    }
+    TempFiles.push_back(PPI.StderrPath.data());
```
rafaelauler, line 310: I think you don't need this else{}, you can just prepend your tmpdir to "perf.script" if it is set and continue to use createTemporaryFile. From what I could read in LLVM libSupport, it will check if it is an absolute path or not, and will only prepend /tmp if it is not an absolute path.

src/DataAggregator.cpp
aagontuk: Oops that was a typo.

src/DataAggregator.cpp
aagontuk: Thank you for the suggestions. I think it won't work. There is an [assertion](https://llvm.org/doxygen/Path_8cpp_source.html#l00835) in `createTemporaryFile()` which only allows valid file names not directories.

rafaelauler: You're right! Then I think it makes sense to just set the default value of opts::TmpDirectory to /tmp and always use your solution that calls createUniqueFile. Also, instead of putting OutFileName in a string, just pass it directly as a Twine, like "sys::fs::createUniqueFile(Twine(opts::TmpDirectory) + "/perf.script-%%%%%.out". 
Also, I can see the indentation looks off in PPI.StdoutPath, for example. BOLT basically uses the same rules as LLVM coding standards, so I would suggest to configure your editor to that codebase. The easiest way to enforce this is to run the code through clang-format, but I think I can fix your code before it is officially merged. Anyway, these are minor things, I'm just trying to make sure your code goes in without the need for any change, otherwise it looks good, thanks!

rafaelauler: I just noticed that the only advantage of calling createTemporaryFile over explicitly specifying "/tmp" like I suggested that we do is that createTemporaryFile actually do support your use case by changing the temporary directory via an env var. Would that work for you? (setting any env var such as "TMPDIR", "TMP", "TEMP", "TEMPDIR")

aagontuk: Setting env var is working. I didn't think about that. Thank you.

PR #103, opened 2020-10-17
Handle TLS relocations on AArch64

yota9:
Some of the TLS relocatios like R_AARCH64_TLSDESC_ADR_PAGE21 must be
handled by bolt and should not be skipped by the removed condition. Some
of the TLS relocations like R_AARCH64_TLS_TPREL64 could really be skipped
here, but AFAIU this condition was added as part of BOLT its self optimization, so
to prevent future problems here my suggestion is not to add another condition
like "isTLS(RType) && isTLSRelocatable(RType)", but just remove it since
absense of this condition should not broke any other TLS relocation.
Vladislav Khmelevsky,

### DIFF
--- a/src/RewriteInstance.cpp
+++ b/src/RewriteInstance.cpp
@@ -2230,8 +2230,8 @@ void RewriteInstance::readRelocations(const SectionRef &Section) {
       RType &= ~ELF::R_X86_64_converted_reloc_bit;
     }
 
-    // No special handling required for TLS relocations.
-    if (Relocation::isTLS(RType))
+    // No special handling required for TLS relocations on X86.
+    if (Relocation::isTLS(RType) && BC->isX86())
       continue;
 
     if (BC->getDynamicRelocationAt(Rel.getOffset())) {


yota9:
I've signed the CLA
Gentle ping @maksfb

rafaelauler:
Hi @yota9, thanks for the patch!

Do you think you would be able to come up with a small test case, such as this one

https://github.com/facebookincubator/BOLT/blob/master/test/X86/section_reloc_with_addend.s

that shows the problem being fixed?

I understand how we would need to indeed process such relocations in AArch64, but I think this is not necessary for X86, so we should add a check for X86 and only bail early if we are in X86, then proceed to handle for AArch64.

I'm not sure if we would handle correctly this operation: Page(G(GTLSDESC(S+A))) – Page(P)  by just looking at the symbol address (I took this expression from AArch64 ELF ABI, document ARM IHI 0056B).  We would need to read the immediate from the ADRP instruction instead of just trying to set the target of the relocation to S + A, similar to what is being done here https://github.com/facebookincubator/BOLT/blob/master/src/RewriteInstance.cpp#L1843  (and the rationale explained in the comments).

Did you verify the TLS reference continues to work after BOLT process it?

yota9:
Dear @rafaelauler, thank you for your comment!

> Did you verify the TLS reference continues to work after BOLT process it?

The TLS works fine after this patch, I've looking the objdump and the addresses are fine too.

> I understand how we would need to indeed process such relocations in AArch64, but I think this is not necessary for X86

I'm not sure about the X86 TLS relocations, I've mostly have an ARM experience. But if you say so I will update the patch to skip TLS relocaitons only for X86 targets.

>  We would need to read the immediate from the ADRP instruction instead of just trying to set the target of the relocation to S + A

Yes, AFAIU analyzeRelication function calls Relocation::extractValue (for AARCH64 binaries) which does it for  R_AARCH64_TLSDESC_ADR_PAGE21 relocation type :)

> Do you think you would be able to come up with a small test case, such as this one

OK I will try :)

Thank you!

yota9:
Gentle ping @rafaelauler @maksfb

rafaelauler:
Hi @yota9! All you said makes sense. I didn't see your update message, sorry. We haven't figured out yet how to perform tests cross-platform. We will likely be moving to the monorepo in the future, and from there we can probably add a dependency on lld, and then run lld to link and produce the input to the test. For now, let's proceed without the test. You can just update your pull request with the check to skip only if we are processing X86 like we discussed.

yota9:
@rafaelauler Thank you! I've updated the patch :)

PR #135, opened 2021-03-24
Ignore optimization passes for function with unsupported instructions

amoksyakov:
Summary:
Reorder-blocks optimization pass doesn't take into account that
available offset for legacy Jcc instructions (for example,
JRCXZ - operand 8 bits) has to be less than 255 bytes.
It's rare case and to exclude such functions with unsupported
instructions from optimization passes added extra checking

Alexey Moksyakov


rafaelauler:
Thanks for the patch! Could you write a testcase? Example: the one in https://github.com/facebookincubator/BOLT/commit/01f471e7108f04b66fd7d3689f3cae3ec8ff6449 in the "test" folder.

maksfb:
If the destination of `JRCXZ` is in a different function the fix wouldn't work, right?

yavtuk:
> If the destination of `JRCXZ` is in a different function the fix wouldn't work, right?
Yes, this fix for the BB reordering inside one function. For instance, 
https://github.com/openssl/openssl/blob/master/crypto/bn/asm/x86_64-mont5.pl#L3319

amoksyakov:
@rafaelauler Hi Rafael, which kind of functionality would you like to cover?

rafaelauler:
The smallest test case you can think of that still triggers the bug, so we keep track of the bug and we don't regress in the future. For this case, you can try setting the edge counts of some branches in a way that will cause bolt to trigger the bug. See the example of https://github.com/facebookincubator/BOLT/blob/master/test/X86/shrinkwrapping-alignment.s to see how we can set the branch weights from the test itself.

rafaelauler:
If you are not familiar with LIT, by the way, a good read is https://llvm.org/docs/TestingGuide.html

yavtuk:
@rafaelauler Ok, I got it, thanks.

amoksyakov:
@rafaelauler  Hello, Rafael, the test is added, sorry for delay

rafaelauler:
@amoksyakov thanks for the test case! The fix looks good for now, but we should probably work in the future to add proper support for this instruction (instead of just bailing on optimizing the function that has one of these). Ideally, we would have a pass that fixes this instruction once we detect that we are exceeding its range for a given layout.
For example, instead of setIgnored(), we would set a different flag to signal that this function needs to relax branches. Then, in a pass before finalization, we check functions that need relaxation and insert an extra unconditional branch in case a JRCXZ has exceeded its range. Let me know if you are interested in working on this. I'll merge this for now.

amoksyakov:
@rafaelauler Yes, I am interested in it. Also I'd like to ask you about functionality adding extra sections to segments. Do you have a pain with it?

rafaelauler:
@amoksyakov awesome. Adding extra sections isn't a problem, but adding new segments (for instrumentation data) isn't supported yet. So we basically limit ourselves to expand the same segment. On the instrumentation front, a larger pain is that it doesn't support instrumenting shared libraries either.

test/X86/bug-reorder-bb-jrcxz.s
```diff
@@ -0,0 +1,597 @@
...
+# REQUIRES: system-linux
+
+# RUN: llvm-mc -filetype=obj -triple x86_64-unknown-unknown \
+# RUN:   %s -o %t.o
+# RUN: link_fdata %s %t.o %t.fdata
+# RUN: %host_cc --version
+# RUN: %host_cc %t.o -falign-labels -march=native -o %t.exe -Wl,-q
+
+# RUN:  llvm-bolt %t.exe -o %t.bolted -data %t.fdata \
+# RUN:    -debug -print-disasm -print-all -v=2 -lite=false \
```
rafaelauler, line 19: We don't use -debug on tests because if we build a version of BOLT without debug_info, it will fail the test.

test/X86/bug-reorder-bb-jrcxz.s
```diff
@@ -0,0 +1,597 @@
...
+# RUN:    -reorder-blocks=cache+ -reorder-functions=hfsort \
+# RUN:    -split-functions=2 -split-all-cold -split-eh -dyno-stats \
+# RUN:    -print-finalized -max-funcs=200 | FileCheck %s
```
rafaelauler, line 22: -print-finalized and -print-disasm are redundant when -print-all is specified.

test/X86/bug-reorder-bb-jrcxz.s
```diff
@@ -0,0 +1,597 @@
...
+
+# CHECK: const: Assertion `isIntN(Size * 8 + 1, Value) && \
```
rafaelauler, line 24: The check should check that it does not assert (that BOLT is indeed fixed). You can do that with "CHECK-NOT", and check that old bolt fails the test, but new bolt passes the test.

But I would rather not check an assertion (because the build can be done without assertions enabled). So it's better to just check that BOLT finishes without issues (don't check anything, and if BOLT crashes, the test will faill because not command line in RUN: is supposed to return non-zero exit code).

rafaelauler (changes requested):
Can you write the diff on top of rebased-bolt branch? That is the most updated version of BOLT and we are currently transitioning to it, and we aren't doing any new commits on top of old main branch.

PR #143, opened 2021-04-23
Summary:

amoksyakov:
Reorder-blocks optimization pass doesn't take into account that
available offset for legacy Jcc instructions (for example,
JRCXZ - operand 8 bits) has to be less than 255 bytes.
It's rare case and to exclude such functions with unsupported
instructions from optimization passes added extra checking

Alexey Moksyakov


amoksyakov:
@rafaelauler I broke the previous PR, sorry, old link https://github.com/facebookincubator/BOLT/pull/135.

bolt/test/X86/bug-reorder-bb-jrcxz.s
```diff
@@ -0,0 +1,643 @@
...
+# https://github.com/openssl/openssl/blob/master/crypto/bn/asm/x86_64-mont5.pl#L3319
+
+# REQUIRES: system-linux
+
+# RUN: llvm-mc -filetype=obj -triple x86_64-unknown-unknown \
+# RUN:   %s -o %t.o
+# RUN: link_fdata %s %t.o %t.fdata
+# RUN: %host_cc %t.o -falign-labels -march=native -o %t.exe -Wl,-q
+
+# RUN:  llvm-bolt %t.exe -o %t.bolted -data %t.fdata \
```
rafaelauler, line 17: Running BOLT on that input is crashing for me with:

/RewriteInstance.cpp:3266: void llvm::bolt::RewriteInstance::mapCodeSections(llvm::RuntimeDyld &): Assertion `TextSection->hasValidSectionID() && ".text section should be valid"' failed.

Run  "ninja check-bolt"  (or "make check-bolt") to see that, after configuring the project with -DLLVM_ENABLE_PROJECTS="bolt" (this is new for bolt in monorepo)

bolt/test/X86/bug-reorder-bb-jrcxz.s
```diff
@@ -0,0 +1,643 @@
...
+# RUN:    -reorder-blocks=cache+ -reorder-functions=hfsort \
+# RUN:    -split-functions=2 -split-all-cold -split-eh -dyno-stats \
+# RUN:    -print-finalized | FileCheck %s
+
+# CHECK-NOT: "Assertion `isIntN(Size * 8 + 1, Value)"
```
rafaelauler, line 22: In case of an assertion firing, the test will fail before "FileCheck" runs because BOLT will crash.

But that is not an issue because I believe that in the rebased BOLT, when BOLT fails to fix up jrcxz, it does not assert anymore, but it prints an error message and continues processing. Then we need to update this message to match that and verify that we are not hitting this path.

Verify that the test fails without your patch and passes after your patch is applied, i.e., after adding the setIgnored() line to BinaryFunction.cpp, the test starts passing but does not pass without it.

rafaelauler (changes requested):
Thanks for rebasing this, Alexey! I have some comments below

amoksyakov:
@rafaelauler hello, the reason for the absence of the assertion firing is that  X86AsmBackend::applyFixup doesn't contain the following assert:
  // Check that uppper bits are either all zeros or all ones.
  // Specifically ignore overflow/underflow as long as the leakage is
  // limited to the lower bits. This is to remain compatible with
  // other assemblers.
  assert(isIntN(Size * 8 + 1, Value) &&
           "Value does not fit in the Fixup field");

I've added this assert to rebased branch and check the test with and without patch. Test works correctly for both cases.

if you aren't going to use this assert then there is the following message in the log, but to see it we have to use -debug option.
< unknown > : 0 : error: value of -2105 is too large for field of 1 byte.
Estimated size for function bn_sqrx8x_internal/1 pre-split is <0x8d2, 0x0>
assembler backend - pre-layout

But during processing bolt is not crashed and I've got a segmentation fault during attemp to run bolted binary.
So, we can apply one of the following:
1) Restore assert(isIntN(Size * 8 + 1, Value) .... or
2) Replace the error message on something else
                   Asm.getContext().reportError(
                                   Fixup.getLoc(), "value of " + Twine(SignedValue) +
                                   " is too large for field of " + Twine(Size) + 
                                   ((Size == 1) ? " byte." : " bytes.")); 

If you get a better idea I can check it 😃

rafaelauler:
Hi Alexey,

I think we should go with checking for "value of -2105 is too large for field of 1 byte" in the logs.  It should not require -debug (it is showing for me without -debug, let me know if you can't see it) and it is also used in other LLVM tests: https://github.com/llvm/llvm-project/blob/main/llvm/test/MC/X86/x86-jcxz-loop-fixup.s 

By the way, I forced push our rebase-branch to a previous commit because I am working on an official rebased version of bolt, and I will soon publish it again, but feel free to rebase your commit on top of our current rebased branch because it shouldn't affect your patch, and then I will personally merge it once it is ready.

rafaelauler:
Oh, another thing that I observed when trying to merge your commit is that it broke this test for us:

X86/fix-branches-jrcxz.s

Did you also see this? This is testing that we process jrcxz correctly, and I think it doesn't make sense after your patch, so we should change that test to XFAIL (expect to fail) and later revisit it once we are properly handling functions with jrcxz.  You can do that by adding this line to fix-branches-jrcxz.s:

XFAIL: *

Ideally after header comments so it's easily visible.

amoksyakov:
@rafaelauler I've added extra one function to fix-branchs-jrcxz test in order to have at least one valid function and therefore we have the valid text section. The test is completed successfully

bolt/src/BinaryFunction.cpp
```diff
@@ -1171,6 +1171,10 @@ bool BinaryFunction::disassemble() {
         const bool IsCondBranch = MIB->isConditionalBranch(Instruction);
         MCSymbol *TargetSymbol = nullptr;
 
+        if (BC.MIB->isUnsupportedBranch(Instruction.getOpcode())) {
+            setIgnored();
```
maksfb, line 1175: We need to check the destination of the branch. If it is a different function, we need to ignore that function too.

bolt/test/X86/bug-reorder-bb-jrcxz.s
amoksyakov: done

bolt/test/X86/bug-reorder-bb-jrcxz.s
amoksyakov: done

bolt/src/BinaryFunction.cpp
amoksyakov: added ignore of a target function

bolt/src/BinaryFunction.cpp
```diff
@@ -1171,6 +1171,16 @@ bool BinaryFunction::disassemble() {
         const bool IsCondBranch = MIB->isConditionalBranch(Instruction);
         MCSymbol *TargetSymbol = nullptr;
 
+        if (BC.MIB->isUnsupportedBranch(Instruction.getOpcode())) {
+          setIgnored();
+          if (!containsAddress(TargetAddress)) {
+            TargetSymbol = getOrCreateLocalLabel(TargetAddress);
+            auto TargetFunc = BC.getFunctionForSymbol(TargetSymbol);
+            if (TargetFunc)
+              TargetFunc->setIgnore();
```
maksfb, line 1180:
```
  if (BinaryFunction *TargetFunc =
          BC.getBinaryFunctionContainingAddress(TargetAddress))
    TargetFunc->setIgnored();

rafaelauler:
Thanks, Alexey! This looks good and we will be merging it into the next rebased branch.

Issue #141, opened 2021-04-20
assert failed when we use perf2bolt


dongAxis:
hello, 

   we occur an failed assert when we try to use the command:
    sudo ./perf2bolt -p ./perf.data ./observer -o test

the crash log is :

 PERF2BOLT: 28426 samples (2.1%) were ignored
 14265 PERF2BOLT: traces mismatching disassembled function contents: 34096 (0.1%)
 14266 PERF2BOLT: out of range traces involving unknown regions: 5375932 (13.1%)
 14267 BOLT-WARNING: function bn_gather5 has an object detected in a padding region at address 0xbae1580
 14268 BOLT-WARNING: function rsaz_avx2_eligible has an object detected in a padding region at address 0xbb02e80
 14269 BOLT-WARNING: function rsaz_512_gather4 has an object detected in a padding region at address 0xbb04a40
 14270 BOLT-WARNING: unable to disassemble instruction at offset 0x1a (address 0xbb1e01a) in function ecp_nistz256_precomputed
 14271 BOLT-WARNING: unable to disassemble instruction at offset 0x11 (address 0xbb6e7d1) in function _vpaes_consts/1
 14272 BOLT-WARNING: function aesni_set_encrypt_key has an object detected in a padding region at address 0xbb72e40
 14273 BOLT-WARNING: unable to disassemble instruction at offset 0x30 (address 0xbb797f0) in function K_XX_XX/1
 14274 BOLT-WARNING: unable to disassemble instruction at offset 0x1 (address 0xbb7cbc1) in function K256/2
 14275 BOLT-WARNING: function Camellia_Ekeygen has an object detected in a padding region at address 0xbb86200
 14276 BOLT-WARNING: function EVP_chacha20_poly1305 has an object detected in a padding region at address 0xbb88b50
 14277 BOLT-WARNING: function aesni_gcm_encrypt has an object detected in a padding region at address 0xbbbc240
 14278 BOLT-WARNING: function gcm_ghash_avx has an object detected in a padding region at address 0xbbc32c0
 14279 BOLT-WARNING: function poly1305_blocks_avx2/1(*2) has an object detected in a padding region at address 0xbbd6180
 14280 BOLT-WARNING: unable to disassemble instruction at offset 0x9b (address 0xbbdf4fb) in function RC4_options
 14281 BOLT-WARNING: unable to disassemble instruction at offset 0x8 (address 0xbbf0448) in function iotas/1
[...]

PR #145, opened 2021-04-27
fix crash when the value of lsda is 0

dongAxis:
[Summary]
We have reported [an issue](https://github.com/facebookincubator/BOLT/issues/141) when bolt to handle some binary that compiled with GCC.

After analyzing, we found that there are an FDE and CIE:
This CIE shows the FDE related to the CIE has LSDA and the encoding of LSDA is DW_EH_PE_pcrel.
But the LDSA value is zero in FDE, so it makes Bolt crash and reports error.

This patch wants to fix that issue.

Author: haishan.fhs & yifeng.dongyifeng


rafaelauler:
Thank for the investigation @dongAxis! I saw your diff in LLVM and I guess you forgot to add the testcase there?

It would also be awesome if you could comment there which GCC version is generating this incorrect FDE, and under which circumstances -- how can we repro that bug, which source code is producing that, just so we know it's not an issue of a broken source code, and that it needs to be fixed at that level (LLVM).

PR #150, opened 2021-05-09
Add BOLT build and test using Github Actions

aaupov:
Add continuous build and testing triggered by GitHub pushes and pull requests.
Builds on x86 ubuntu-latest with system gcc and clang


rafaelauler:
That's awesome, thanks! Is it ready?

Issue #130, opened 2021-03-16
Expand auto types


MaskRay:
https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable `auto` appears to be aggressively used. The code should be refactored before dropping into the upstream.

steveire:
Note that at least flang and mlir use `auto` a lot too, so there is plenty of precedent for using it both inside and outside llvm. Any implication in the coding standard that `auto` shouldn't be used normally is probably just out of date.

aaupov:
@MaskRay, we've revisited the use of `auto` in the codebase and reduced it to lambdas, iterators, template-derived types, places where type is explicitly given in the context, and alias variables. There might still be some places with `auto` where changing it to the expected type changed the binary, so we've left them out.
Commit 684c48d60863d4712b497f8b51ba04421f64ebc8

PR #153, opened 2021-05-11
Introduce LoopInversion optimization pass

yota9:
This patch inroduces LoopInversion pass. It's main purpose to ensure
that the loop layout is optimal depending on the profile information. So
if profile information shows that the loop is used, the uncoditional
jump instruction (if any) must be executed only once and vice-versa. Please take
a look to the pass header file and test for more details.

Also change link_fdata script a bit, to be able to change FDATA prefix,
like the FileCheck does.

Vladislav Khmelevsky,


yota9:
Code cleanup & clang-format updates

bolt/src/Passes/ReorderUncondJump.h
```diff
@@ -0,0 +1,59 @@
...
+//===----------------------------------------------------------------------===//
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LLVM_TOOLS_LLVM_BOLT_PASSES_REORDERUNCONDJUMP_H
+#define LLVM_TOOLS_LLVM_BOLT_PASSES_REORDERUNCONDJUMP_H
+
+#include "BinaryPasses.h"
+
+/*
```
rafaelauler, line 17: LLVM coding standards require us to use only C++-style comments  (//)

bolt/src/Passes/BinaryPasses.cpp
```diff
@@ -444,6 +448,78 @@ void ReorderBasicBlocks::modifyFunctionLayout(BinaryFunction &BF,
...
+  bool IsChanged = false;
+  if (BF.layout_size() < 3 || !BF.hasValidProfile())
+    return false;
+
+  BF.updateLayoutIndices();
+  for (auto BB : BF.layout()) {
+    if (BB->succ_size() == 1 && BB->pred_size() == 1) {
+      auto SuccBB = *BB->succ_begin();
+      auto PredBB = *BB->pred_begin();
+      const auto BBIndex = BB->getLayoutIndex();
```
rafaelauler, line 461: `const unsigned BBIndex`

PS: See #130

bolt/src/Passes/BinaryPasses.cpp
```diff
@@ -444,6 +448,78 @@ void ReorderBasicBlocks::modifyFunctionLayout(BinaryFunction &BF,
...
 
+bool ReorderUncondJump::runOnFunction(BinaryFunction &BF) {
+  bool IsChanged = false;
+  if (BF.layout_size() < 3 || !BF.hasValidProfile())
+    return false;
+
+  BF.updateLayoutIndices();
+  for (auto BB : BF.layout()) {
+    if (BB->succ_size() == 1 && BB->pred_size() == 1) {
+      auto SuccBB = *BB->succ_begin();
```
rafaelauler, line 459: `BinaryBasicBlock *`

bolt/src/Passes/BinaryPasses.cpp
```diff
@@ -444,6 +448,78 @@ void ReorderBasicBlocks::modifyFunctionLayout(BinaryFunction &BF,
...
+      auto PredBB = *BB->pred_begin();
+      const auto BBIndex = BB->getLayoutIndex();
+      const auto SuccBBIndex = SuccBB->getLayoutIndex();
+      if (SuccBB == PredBB && BB != SuccBB && BBIndex != 0 &&
+          SuccBBIndex != 0 && SuccBB->succ_size() == 2 &&
+          BB->isCold() == SuccBB->isCold()) {
+
+        /* Get the second successor (after loop BB) */
+        BinaryBasicBlock *OtherSucc = nullptr;
+        for (auto Succ : SuccBB->successors())
```
rafaelauler, line 469: Wrap loop body with {}

bolt/src/Passes/BinaryPasses.cpp
```diff
@@ -184,6 +184,10 @@ ReorderBlocks("reorder-blocks",
   cl::ZeroOrMore,
   cl::cat(BoltOptCategory));
 
+static cl::opt<bool> NoUncondJumpReorder(
```
rafaelauler, line 187: Can we flip this to be "UncondJumpReorder" and start with default "false" (pass disabled) until we test it and make sure it is mature, then later we can flip to default "true"?

rafaelauler (changes requested):
Thanks @yota9! A few nits below. If this is to be enabled by default, we probably should later take a closer look at the impact of this pass to gauge pros and cons (processing time increase vs. wins), but either way it's nice to see submissions like this.

yota9:
@rafaelauler Thank you for review! I've resolved all your comments. But I still hope to activate this pass by default one day.. Thank you!

bolt/src/Passes/ReorderUncondJump.h
```diff
@@ -0,0 +1,57 @@
...
+// the unconditional jump is only used once on the loop start.
+
+namespace llvm {
+namespace bolt {
+
+class ReorderUncondJump : public BinaryFunctionPass {
+public:
+  explicit ReorderUncondJump() : BinaryFunctionPass(false) {}
+
+  const char *getName() const override { return "ReorderUncondJump"; }
```
rafaelauler, line 47: Here, instead of "ReorderUncondJump", we should use the flag name used to control this pass execution, thus, "uncond-jmp-reorder".

bolt/src/Passes/BinaryPasses.cpp
```diff
@@ -184,6 +184,11 @@ ReorderBlocks("reorder-blocks",
   cl::ZeroOrMore,
   cl::cat(BoltOptCategory));
 
+static cl::opt<bool>
+    UncondJumpReorder("uncond-jmp-reorder",
+                      cl::desc("reorder loops unconditional jump"),
+                      cl::init(false), cl::cat(BoltCategory), cl::ReallyHidden);
```
rafaelauler, line 190: Go ahead and change it to "true".

bolt/src/Passes/BinaryPasses.cpp
```diff
@@ -444,6 +449,79 @@ void ReorderBasicBlocks::modifyFunctionLayout(BinaryFunction &BF,
...
+    if (BB->succ_size() == 1 && BB->pred_size() == 1) {
+      BinaryBasicBlock *SuccBB = *BB->succ_begin();
+      BinaryBasicBlock *PredBB = *BB->pred_begin();
+      const unsigned BBIndex = BB->getLayoutIndex();
+      const unsigned SuccBBIndex = SuccBB->getLayoutIndex();
+      if (SuccBB == PredBB && BB != SuccBB && BBIndex != 0 &&
+          SuccBBIndex != 0 && SuccBB->succ_size() == 2 &&
+          BB->isCold() == SuccBB->isCold()) {
+
+        /* Get the second successor (after loop BB) */
```
rafaelauler, line 468: Change to //

bolt/src/Passes/ReorderUncondJump.h
```diff
@@ -0,0 +1,57 @@
...
+//===----------------------------------------------------------------------===//
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LLVM_TOOLS_LLVM_BOLT_PASSES_REORDERUNCONDJUMP_H
+#define LLVM_TOOLS_LLVM_BOLT_PASSES_REORDERUNCONDJUMP_H
+
+#include "BinaryPasses.h"
+
+// This pass founds casses when BBs have layout:
```
rafaelauler, line 17: s/casses/cases

bolt/test/X86/reorder-uncond-jmp.s
```diff
@@ -0,0 +1,38 @@
+# REQUIRES: system-linux
+
+# RUN: llvm-mc -filetype=obj -triple x86_64-unknown-unknown \
+# RUN:   %s -o %t.o
+# RUN: link_fdata %s %t.o %t.fdata
+# RUN: link_fdata %s %t.o %t.fdata2 "FDATA2"
+# RUN: %host_cc %cflags %t.o -o %t.exe -Wl,-q
+# RUN: llvm-bolt %t.exe -data %t.fdata -reorder-blocks=cache+ -print-finalized \
+# RUN:    -uncond-jmp-reorder -o %t.out | FileCheck %s
```
rafaelauler, line 9: Even when you set the pass to run by default, leave the flag here, so it's explicit that we are testing this pass.

rafaelauler (commented):
Hi @yota9, I finished evaluating it with -time-opts and the overhead is negligible (see below). Yet, the transformation itself seems to be strictly improving the layout of loops and based on the impact that you observed, I think it makes sense to make it ON by default.

===-------------------------------------------------------------------------===
                          Binary Function Pass Manager
===-------------------------------------------------------------------------===
  Total Execution Time: 286.4311 seconds (51.3369 wall clock)

   ---User Time---   --System Time--   --User+System--   ---Wall Time---  --- Name ---
  16.9790 (  7.1%)   0.6428 (  1.3%)  17.6219 (  6.2%)  17.4563 ( 34.0%)  indirect-call-promotion
  140.0243 ( 58.7%)   1.1766 (  2.5%)  141.2010 ( 49.3%)  13.5946 ( 26.5%)  identical-code-folding
   3.1578 (  1.3%)   0.0371 (  0.1%)   3.1950 (  1.1%)   3.1924 (  6.2%)  fix-branches
   5.5537 (  2.3%)   8.4021 ( 17.5%)  13.9558 (  4.9%)   1.8580 (  3.6%)  finalize-functions
   1.7536 (  0.7%)   0.0363 (  0.1%)   1.7899 (  0.6%)   1.7868 (  3.5%)  reorder-functions
   1.6080 (  0.7%)   0.0000 (  0.0%)   1.6080 (  0.6%)   1.6031 (  3.1%)  PLT call optimization
   1.5129 (  0.6%)   0.0000 (  0.0%)   1.5129 (  0.5%)   1.5130 (  2.9%)  validate-internal-calls
   1.3827 (  0.6%)   0.0269 (  0.1%)   1.4097 (  0.5%)   1.3988 (  2.7%)  lower-annotations
  29.2087 ( 12.2%)  37.6194 ( 78.4%)  66.8281 ( 23.3%)   1.3941 (  2.7%)  split-functions
   1.3091 (  0.5%)   0.0008 (  0.0%)   1.3099 (  0.5%)   1.3071 (  2.5%)  simplify-conditional-tail-calls
   1.1957 (  0.5%)   0.0000 (  0.0%)   1.1957 (  0.4%)   1.1943 (  2.3%)  eliminate-unreachable
   1.1884 (  0.5%)   0.0000 (  0.0%)   1.1884 (  0.4%)   1.1885 (  2.3%)  simplify-read-only-loads
   0.8007 (  0.3%)   0.0000 (  0.0%)   0.8007 (  0.3%)   0.8005 (  1.6%)  inlining
   0.7069 (  0.3%)   0.0000 (  0.0%)   0.7069 (  0.2%)   0.7063 (  1.4%)  inst-lowering
   0.4962 (  0.2%)   0.0000 (  0.0%)   0.4962 (  0.2%)   0.4962 (  1.0%)  strip-rep-ret
   0.4656 (  0.2%)   0.0020 (  0.0%)   0.4675 (  0.2%)   0.4630 (  0.9%)  specialize-memcpy
  11.2266 (  4.7%)   0.0000 (  0.0%)  11.2266 (  3.9%)   0.4392 (  0.9%)  reorder-blocks
   0.3920 (  0.2%)   0.0015 (  0.0%)   0.3935 (  0.1%)   0.3912 (  0.8%)  print dyno-stats after optimizations
  19.2189 (  8.1%)   0.0000 (  0.0%)  19.2189 (  6.7%)   0.3633 (  0.7%)  aligner
   0.1048 (  0.0%)   0.0000 (  0.0%)   0.1048 (  0.0%)   0.1048 (  0.2%)  assign-sections
   0.0356 (  0.0%)   0.0000 (  0.0%)   0.0356 (  0.0%)   0.0356 (  0.1%)  print-stats
   0.1330 (  0.1%)   0.0115 (  0.0%)   0.1445 (  0.1%)   0.0298 (  0.1%)  ReorderUncondJump <===
   0.0199 (  0.0%)   0.0000 (  0.0%)   0.0199 (  0.0%)   0.0199 (  0.0%)  patch-entries
   0.0000 (  0.0%)   0.0000 (  0.0%)   0.0000 (  0.0%)   0.0000 (  0.0%)  frame-optimizer
   0.0000 (  0.0%)   0.0000 (  0.0%)   0.0000 (  0.0%)   0.0000 (  0.0%)  peepholes
   0.0000 (  0.0%)   0.0000 (  0.0%)   0.0000 (  0.0%)   0.0000 (  0.0%)  reorder-data
   0.0000 (  0.0%)   0.0000 (  0.0%)   0.0000 (  0.0%)   0.0000 (  0.0%)  retpoline-insertion
   0.0000 (  0.0%)   0.0000 (  0.0%)   0.0000 (  0.0%)   0.0000 (  0.0%)  alloc-combiner
  238.4740 (100.0%)  47.9572 (100.0%)  286.4311 (100.0%)  51.3369 (100.0%)  Total

yota9:
@rafaelauler Thank you, done!

rafaelauler:
FYI I already imported this diff to our internal branch and this is going through a round of tests using our CI. If everything goes well, it should be merged in the near future.

yota9:
@rafaelauler Thank you, waiting for the results :)

maksfb:
@yota9, could you share on what architecture you see the improvement and how the profile is collected?

bolt/src/Passes/ReorderUncondJump.h
```diff
@@ -0,0 +1,57 @@
...
+
+#ifndef LLVM_TOOLS_LLVM_BOLT_PASSES_REORDERUNCONDJUMP_H
+#define LLVM_TOOLS_LLVM_BOLT_PASSES_REORDERUNCONDJUMP_H
+
+#include "BinaryPasses.h"
+
+// This pass founds cases when BBs have layout:
+// #BB1:
+// cmp
+// cond_jmp out
```
maksfb, line 20: nit: `cond_jmp #BB3` to be consistent.

bolt/src/Passes/ReorderUncondJump.h
```diff
@@ -0,0 +1,57 @@
...
+// #BB2:
+// <code>
+// jmp #BB1
+// #BB3 <out>
+//
+// And swaps BB1 and BB2:
```
maksfb, line 26: Could you keep BB #'s the same as in the original (the way BOLT does it)? You'll have to introduce BB0.

maksfb (commented):
Could you move the pass to a separate .cpp file?

bolt/src/Passes/ReorderUncondJump.h
yota9: Sorry, you mean to change BB1 to BB0 & etc?

maksfb: Sorry, I mean treat BB#'s as BB names/labels, i.e. after the pass:

#BB0:
...
jmp #BB1
#BB2:
<loop body>
#BB1:
cmp ...
jncc #BB2
#BB3:
<loop exit>

yota9:
@maksfb Currently I've checked it on x86. 
I've wrote this patch some time ago, I think from this time we've tested it with both instrumentation and perf.
You can check the results on something as trivial as:

volatile int a = 5;
void foo(void) {
  for (int i = 0; i < 200000; ++i)
      a = a*a;
} 

compiled with -O0

yota9:
@maksfb Thank you for your comments, done!

yota9:
Hello!

maksfb (approved):
Overall looks good!

maksfb:
I ran the microbenchmark on Haswell CPU with increased loop trip count. I couldn't spot any performance difference. We also haven't noticed difference on a couple of larger benchmarks.

maksfb:
Loop inversion optimization is more canonical name for the transformation. Do you think it makes sense to rename the pass?

yota9:
@maksfb Thank you for your comments!
I will rename it, I'm bad at naming something :)
About benchmark. Did you tried to compile it with O0? I've investigated the gcc asm, it seems to be the problem is ftree-ch option, for some reason it duplicates the conditions before the loop instead of just jump on the loop conditions below. Let me show this.
Loop with -03:
`  1050:       39 d6                   cmp    %edx,%esi
    1052:       7e 20                   jle    1074 <main+0x34>
    <loop body>
    1070:       39 d6                   cmp    %edx,%esi
    1072:       75 e4                   jne    1058 <main+0x18>
    1074:       8b 05 9a 2f 00 00       mov    0x2f9a(%rip),%eax        # 4014 <a>
`
And loop with -03 -fno-tree-ch:
`
   1050:       eb 18                   jmp    106a <main+0x2a>
    <loop body>
    1070:       39 f2                   cmp    %esi,%edx
    1072:       7c e4                   jl     1058 <main+0x18>
`
Which is kind of strange  for my taste.
So my сompiler is creating the second variant by default , but the default gcc with optimization option higher then -O0 will produce the first variant, which I think won't be optimized by this pass..

maksfb:
@yota9, right, compiled with -O0 and I've check the assembly code. A can see 2x branch count reduction with `perf -e branches ....`, but no CPU time change.

yota9:
@maksfb This is probably due to branch predictor. Our minimum test is something like client-server messenger with a bit of a logic there, maybe branch predictor buffer could not handle the whole application in it..  It is just my guess, but you see 2x branch reduction and I presume it must give performance on the real-world applications. So for example with this client-server test (it is extremely bad test, but anyway) I see 1/3 of fewer messages processed by constant time without this patch :)

yota9:
@maksfb  Thank you for your comments, done!

rafaelauler:
If @maksfb has no further comments, I'll import this and commit. Let me know.

bolt/src/Passes/LoopInversionPass.cpp
```diff
@@ -0,0 +1,103 @@
...
+#include "ParallelUtilities.h"
+
+using namespace llvm;
+
+namespace opts {
+extern cl::OptionCategory BoltCategory;
+
+extern cl::opt<bolt::ReorderBasicBlocks::LayoutType> ReorderBlocks;
+
+static cl::opt<bool> UncondJumpReorder(
```
maksfb, line 21: Please update the option name as well.

maksfb (approved):
Please change the name of the commit and update the summary to reflect the change in the name. After that it's good to go.

yota9:
@maksfb Hello! Do I need to change something else in commit description or is it good to go? Thank you!

maksfb:
It looks good to me.

PR #155, opened 2021-05-14
Docker

aaupov:


maksfb (approved):
Nice!

maksfb (commented):
Could you add linking with jemalloc?

aaupov:
@maksfb:
> Could you add linking with jemalloc?

It's linking with jemalloc:
>-DCMAKE_EXE_LINKER_FLAGS="-ljemalloc_pic -lpthread" 
I've removed -whole-archive directive as it was failing to build. 

I think overall it's a better idea to build docker image and test BOLT inside it. I'll drop PR with Github Actions ubuntu workflow as it's redundant and doesn't produce an artifact.

maksfb:
Sounds great. My comment was specific to `.github/workflows/ubuntu.yml`, but it's not obvious from GH interface.

yota9:
@aaupov 
>  (I've removed -whole-archive directive as it was failing to build)

Is there a possibility, that the linker error gone because it jemalloc is not linked any more? Do you have jemalloc symbols inside llvm-bolt executable? The CMAKE_EXE_LINKER_FLAGS prepends the passed flags to the linker. Thus the jemalloc does not resolve any dependencies and it is just not included in the final binary.  The whole-archive is kind of hack here to prevent this and force to use the library.
I assume that the failing problem is caused because of cmake compiler tests, so the whole jemalloc dep libs must be passed -DCMAKE_EXE_LINKER_FLAGS="-Wl,--whole-archive ljemalloc_pic -Wl,--no-whole-archive -lstdc++ -lpthread -ldl -lm". But here we may have a problem with lpthread passed to linker flags, shared library will have undefined references to pthread symbols (the llvm rules uses CMAKE_THREAD_LIBS_INIT to set LLVM_PTHREAD_LIB variable, which will be empty). Is it a bug? I don't know, prefer make over cmake :)) and I don't know an easy way to pass lpthread to them (CMAKE_SHARED_LINKER_FLAGS has the same problem of flags to be prepended). So as a hack you can pass -DCMAKE_HAVE_LIBC_PTHREAD:INTERNAL=""  flag.

aaupov:
@yota9: you're right, it's not linking with jemalloc if we only pass `-ljemalloc_pic -lpthread`. Still looking at how to cleanly link with jemalloc

aaupov:
Now it's built with jemalloc:
```
$ docker build . --file ~/bolt-repo/aaupov-bolt/.github/workflows/Dockerfile
...
 => => writing image sha256:80c86a7c07e4f8e2eb9170d76056a60b58a0fd1c000cd53386fda446ed4948cd                                                       0.0s
$ docker run -it 80c86a7c0 /bin/bash
root@6bf0ebcb59aa:/# MALLOC_CONF=invalid_flag:foo llvm-bolt --version
<jemalloc>: Invalid conf pair: invalid_flag:foo
```

yota9:
@aaupov Interesting that you don't need  HAVE_LIBC_PTHREAD hack, probably I've got too old cmake :)

PR #156, opened 2021-05-14
Instrumentation: Emit paddings to preserve data alignment

yota9:
Preserve alignment of emitting data type.
Vladislav Khmelevsky,


yota9:
Gentle ping to @rafaelauler @maksfb

rafaelauler (commented):
Hi @yota9-huawei thanks for this patch!

This looks good to me. In upstream LLVM, they fixed some older MCStreamer member functions casing (from EmitBla to emitBla). Lambdas in general should follow the casing used for function names, so could you change EmitX to emitX in this patch?  (I'm going for compliance with https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly )

yota9:
@rafaelauler Sure, done!
Thank you!

rafaelauler:
Thanks! Let me import* this.

* I think you are probably aware of this, but for the sake of transparency, let me briefly explain our current repo organization that will stay like this until we can live in the LLVM repo.

We have an internal source of truth git repo that is mirrored here on github. So every commit we post internally is synced here. Unfortunately commit hashes change because we plan to be constantly rebasing on top of the LLVM main branch. But that is the official BOLT release, by the way. It is completely ported to work in the LLVM monorepo and we don't plan on using the old main branch anymore.

When we receive a github PR, we can't actually accept the push request in the usual github flow because what we have here is just a mirror repo. That's when the import happens, I run a script that cherry picks your work and put in our internal source of truth repo. The advantages of having the patch in that internal repo is that we will run our CI tests on it. These will start testing BOLT with very large binaries that are awkward to live directly in the open-source repo. We currently have about 200+ of such tests. Our goal is to eventually port most of these tests to be open-source tests.

If the patch breaks a test and I know how to fix it, I'll probably just fix them. So sometimes it takes me a bit of time to update the tests. If there is a real issue, I'll try my best to communicate the issue. So that's what happened with the previous diff that introduced a new "default ON" pass in the pipeline. Because it runs by default, it broke a handful of tests and it took some time to update the tests (some tests check that stats remain equal across changes so we keep a close watch on any intended or non-intended changes to the binary). But it should be mostly resolved by now.

This patch in particular passes all tests, and looks reasonable to me, so it should be good.

yota9:
@rafaelauler Thanks for clarifying. Hope to see BOLT in LLVM infrastructure officially soon!

rafaelauler:
Regarding the performance benefits of the previous pass: I ran a bunch of experiments with different techniques to detect changes in CPU time, and I also fixed the internal tests that were checking branch statistics so I was aware of how much it was affecting some traces.

In some binaries, we observe a decrease in total branches. In others, an increase. Regarding CPU time improvements, some experiments that I ran aren't super accurate, so they are unable to detect performance differences smaller than 1% (which are still quite significant). For other experiments, they are able to detect perhaps 0.5% more reliably, but are running a different workload. Overall I didn't detect CPU time differences, but that doesn't mean the pass isn't helping. That's why Maksim was asking for some clarification.

yota9:
@rafaelauler I see. Well it is king of strange that any of this tests does not show the CPU time difference :( I said we've got a test with a counting cycle, where I've seen the huge degradation (33%) in it. Also my colleague tried that pass with this tests https://github.com/json-iterator/go-benchmark and we saw ~0.7% of extra performance boost with this pass.. If it is a problem I will try to find some kind of C benchmark to confirm this results :)
Also it might be CPU specific in theory (however I'm quite sure that jump is quite a "heavy" instruction, and their elimination often gives a performance boost).
BTW I have an interesting story about CPU. Some time ago we've updated ubuntu on one of our servers. After that happened any BOLTed application showed significantly performance decrease. After some time I've found out that with the ubuntu update Intel CPU microcode update was loaded, that was probably fixed meltdown/spectre vulnerabilities. After the microcode update was removed the performance came back to normal as it was before. I don't think that it the reason why you don't see CPU time improvements, but just a story how CPU and it's firmware could influence on the results :)

rafaelauler:
Ohhhh, I see, interesting. Indeed, the meltdown/spectre microcode is a perf killer, so I expect there will be opportunities to improve perf. We have a pass (--insert-retpolines) that inserts code to perform meltdown/spectre mitigation in software via retpoulines, and because we do it with profile information, we also reorder the blocks and branches accordingly, and we are able to reduce the performance overhead of the retpoulines introduced as a result of the mitigation (I know compilers also have a pass to do that). But what you report makes sense, BOLT has a very tight relationship with the specific CPU uarch, and the older is the CPU, typically the larger are the wins, because older CPUs usually suffer with lower cache performance (icache, iTLB and branch prediction performance). So it's entirely possible that a newer uarch comes and renders an optimization perf neutral, but that optimization is still relevant for other chips.

PR #166, opened 2021-06-04
Aarch64: Patch relative relocations

yota9:
The relative relocations must be patched on ARM the same way they patched on X86.
Vladislav Khmelevsky,


rafaelauler:
Thanks @yota9! Is there an easy way to write a test case?

yota9:
@rafaelauler Well the test is self might be any DYN executable for aarch64, which will have relative relocations. But since currently test system is X86-only I don't see any way to add this.. 
Thank you!

PR #167, opened 2021-06-04
Instrumentation: Disable signals on mutex lock

yota9:
When indirect call is instrumented we're locking SimpleHashTable's mutex on get() call.
If while locked we will receive a signal and signal handler also will call indirect function we will end up with deadlock. 
So to avoid such situation let's block all signals during mutex is locked.
Vladislav Khmelevsky,


yota9:
I didn't test it on apple, sorry

rafaelauler:
Good catch! Thanks for fixing this

rafaelauler:
Failed to build on my machine:

llvm-project/bolt/runtime/common.h:445:46: error: use of undeclared identifier 'NULL'
      __sigprocmask(SIG_SETMASK, SignalMask, NULL);
                                             ^

yota9:
@rafaelauler Seems to be you have mac, stddef is not connected for it. 
I will replace it to "(void *)0" now, I think it's ok :)

bolt/runtime/common.h
```diff
@@ -407,13 +431,19 @@ class Mutex {
...
   Lock(Mutex &M) : M(M) {
+    __sigprocmask(SIG_BLOCK, MaskAllSignals, SignalMask);
     while (!M.acquire()) {
     }
   }
-  ~Lock() { M.release(); }
+
+  ~Lock() {
+      M.release();
+      __sigprocmask(SIG_SETMASK, SignalMask, (void *)0);
```
rafaelauler, line 445: Well, this is a C++ file, so I guess we can just use:
`  __sigprocmask(SIG_SETMASK, SignalMask, nullptr);`
instead of going with the C dialect of casting NULL or 0.

rafaelauler:
I actually tested on a Linux machine. Maybe my C standard headers are busted somehow and missing NULL.

bolt/runtime/common.h
yota9: Sure, I don't mind :)

PR #173, opened 2021-06-11
Fdata: Add quotes to function names

yota9:
This patch is part of preparation for golang support. The golang symbols
might have spaces in the name (for example "type..eq.[10]interface {}"). 
Since fdata uses spaces as field separator such names brakes
fdata format, so I suggest to use quotes for symbol names to avoid such errors.

Vladislav Khmelevsky,


bolt/src/DataReader.cpp
```diff
@@ -945,7 +955,10 @@ ErrorOr<Location> DataReader::parseLocation(char EndChar,
   consumeAllRemainingFS();
 
   // Read the string containing the symbol or the DSO name
-  ErrorOr<StringRef> NameRes = parseString(FieldSeparator);
+  if (!expectAndConsumeNS())
```
yota9, line 958: If it is necessary we can remove this error and  use parseString(NameSeparator) if NameSeparator was found and FieldSeparator if it was not. But if it possible for for you @rafaelauler to replace old fdata you use without quotes with new one - I suggest just to use new variant.

yota9:
The old fdata files might be fixed using simple awk script:
`awk '{$2 = "\""$2"\""; $5 = "\""$5"\"";print}' prof.fdata`

maksfb:
Is it possible to include double quotes in the real symbol name as well? I've tried to find if a name with double quotes would constitute a valid name in ELF and couldn't find any rule against it. Plus, we don't want to limit ourselves to ELF. In such case, we should support names with double quotes as well as space. Escaping space in a symbol name with a backslash might be the most straightforward way to add the extend the fdata format. We will also get close to full backwards compatibility automatically.

Is the slash character ("/") in the example you gave ("type..eq.[10]interface {}/") part of the valid golang symbol name? If that's the case, we might need to redesign the way we disambiguate local symbols.

yota9:
>> we should support names with double quotes as well as space. 
I agree, although
>> Escaping space in a symbol name with a backslash might be the most straightforward way to add the extend the fdata format.
Could you please add more details, I don't quite understand how the new format will look..

About the slash character - I will double check it soon. However, I don't know limitations that could forbid the usage of slash character in the name :)

maksfb:
> Could you please add more details, I don't quite understand how the new format will look..

The only extension/fix is the support for escaping characters in the name with backslash. I.e. `function\ name`, would be properly handled for golang.

yota9:
@maksfb I've checked, the slash at the end was redundant.
>> The only extension/fix is the support for escaping characters in the name with backslash

I see, we will handle this. Thank you!

Issue #181, opened 2021-06-28
inline-debug-info test is failing


yota9:
Hello! 
The inline-debug-info is currently failing on linux, seems to be call address is screwed after inlining.
Thank you!

aaupov:
We're aware of the issue. Ubuntu uses Intel CET-enabled PLT section (.plt.sec) which messes up the disassembly of `call puts`. @maksfb was looking at it.

aaupov:
Should be resolved now. The test passes with 403b668

Issue #117, opened 2021-02-16
Disassemble issue with gcc builit binary


raghesh:
Getting the following error while using llvm-bolt with gcc built binary.

"BOLT-ERROR: unable to disassemble instruction in PLT section .plt.got at offset 0x8"

**Tested program - test.c**

int main() {
}

**Commands used**

gcc -O0  -g -Wl,-emit-relocs test.c -o test.gcc.out
llvm-bolt -lite=0 -print-cfg ./test.gcc.out -o /dev/null

**Objdump output**

The output of objdump -d test.gcc.out -j .plt.got

test.gcc.out:     file format elf64-x86-64

Disassembly of section .plt.got:

0000000000001030 <__cxa_finalize@plt>:
    1030:       f3 0f 1e fa             endbr64 
    1034:       f2 ff 25 bd 2f 00 00    bnd jmpq *0x2fbd(%rip)        # 3ff8 <__cxa_finalize@GLIBC_2.2.5>
    103b:       0f 1f 44 00 00          nopl   0x0(%rax,%rax,1)

Looks like the bnd prefix is not recognized. This was tested on an AMD machine.

I tested on a different (intel) machine, the issue is not there. The output of objdump there was as follows.

test.gcc.out:     file format elf64-x86-64

Disassembly of section .plt.got:

00000000000004e0 <__cxa_finalize@plt>:
 4e0:   ff 25 12 0b 20 00       jmpq   *0x200b12(%rip)        # 200ff8 <__cxa_finalize@GLIBC_2.2.5>
 4e6:   66 90                   xchg   %ax,%ax

The bnd prefix is not present here.

aagontuk:
Which gcc version are you using?

>NOTE: BOLT is currently incompatible with the -freorder-blocks-and-partition compiler option. Since GCC8 enables this option by default, you have to explicitly disable it by adding -fno-reorder-blocks-and-partition flag if you are compiling with GCC8.

Did you try compiling with `-fno-reorder-blocks-and-partition`?

raghesh:
> Which gcc version are you using?
> 

I am using gcc-10

> > NOTE: BOLT is currently incompatible with the -freorder-blocks-and-partition compiler option. Since GCC8 enables this option by default, you have to explicitly disable it by adding -fno-reorder-blocks-and-partition flag if you are compiling with GCC8.
> 
> Did you try compiling with `-fno-reorder-blocks-and-partition`?

[...]

PR #184, opened 2021-07-04
Instrumentation support for PIE & shared libs on x86_64 Linux

### LINKED ISSUE #137: Using Bolt in dynamic libary, occur Segmentation fault
#1、 demo
image

the above is the small case, the hello.c is so, the version of gcc is 7.3
 gcc hello.c -o libhello.so -I../include -fPIC -shared           

gcc main.c  -o main -I../include -L../src -lhello

./main 

image

# 2、 Using llvm bolt instrument binary, 
./../../../../build/bin/llvm-bolt ./libhello.so -instrument -instrumentation-file=./hello.fdata -use-old-text -instrumentation-sleep-time=30 -o ./libhello.inst

mv libhello.inst libhello.so
./main 

image

the so occur segment fault, can anyone know the reason? is current bolt dont support dynamic instrument ?

vleonen:
This PR enables instrumentation support for PIE executables, shared libraries, and initial support for statically linked executables on x86_64 Linux systems.
We are using relative jumps in trampoline functions to avoid usage of absolute addresses calculated during build time on the linking stage for initialization/finalization functions and instrumentation calls. These absolute addresses are not valid for randomized addresses for PIE & shared libraries.
Also for correct operation for instrumented shared libraries, we added a set of syscalls and functions for searching a path to the current binary in /proc/self/map_files directory based on the current virtual address and symlinks in this directory. In case/proc/self/map_files directory access is restricted (e.g. in container environment), we added -instrumentation-binpath option to pass information about expected path to an instrumented binary during runtime.

Co-authored-by: Vladislav Khmelevsky <vladislav.khmelevskyi@huawei.com>
Co-authored-by: Elvina Yakubova <elvina.yakubova@huawei.com>

Fixes #137 

Vasily Leonenko,


rafaelauler:
Thanks a lot for working on this!

I pulled the branch, but it is failing two tests in my local machine:

Failed Tests (2):
  BOLT :: X86/instrumentation-pie.c
  BOLT :: X86/instrumentation-shlib.c

Can you verify that?

I see the tests are introduced in the first commit "Tests: add instrumentation tests for PIE exec & shared libs". Usually we add tests in the same commit that introduces a functionality, instead of before. The reason for that is that we want to preserve our ability to bisect problems, and we can't have a given commit in the history that introduces tests that are failing. So I suggest adding the tests together with the first commit that makes them pass. If they were made by different authors, then it's fine to commit then separately, as long as the test commit comes after the functionality.

vleonen:
Hello @rafaelauler
Thank you for your comments for this PR and for checking it.
As complete instrumentation support was implemented in several commits from several authors, I rewrote my branch to put tests on top of it.

As for failed tests - as I can see from Docker CI log - the test failed with an bolt-error mentioned in issues #117 & #181. I reproduced the same behavior on Ubuntu-20.04 with gcc-9. On Ubuntu-18.04 with gcc-7 - there are no such issues.

As I understood from #181 - Bolt team is already working on resolving the root cause of the issue.

Until then we can consider using a workaround commit added on top of my branch (Tests: temp fix for failing tests on newer gcc versions). With this fix - there are no failed BOLT tests on Ubuntu-20.04 with gcc-9. However, more tests will fail with older gcc versions, because "-fcf-protection" option was introduced in gcc-8. So feel free to cherry-pick all commits except the last one with a workaround. I'll appreciate your suggestions on how to handle this gcc version-specific behavior more properly.

vleonen:
Rebased on top of current facebookincubator:rebased branch

rafaelauler:
Thanks Vasily! Overall the diffs look pretty nicely written. But before I go over them, I'm still trying to figure out what is happening with testing to check that things are working as intended. I'm getting:

At X86/instrumentation-pie.c,

`clang  -no-pie -fcf-protection=none bolt/test/X86/instrumentation-pie.c -o build/tools/bolt/test/X86/Output/instrumentation-pie.c.tmp.exe -Wl,-q -pie -fpie`

`bin/llvm-bolt bolt/test/X86/Output/instrumentation-pie.c.tmp.exe -instrument -instrumentation-file=test/X86/Output/instrumentation-pie.c.tmp.fdata    -o test/X86/Output/instrumentation-pie.c.tmp.instrumented`
Output:
> BOLT-INFO: shared object or position-independent executable detected
> BOLT-INFO: Target architecture: x86_64
> BOLT-INFO: BOLT version: b9ce0d79abe7daff171e7c31c18dcec3bf058c82
> BOLT-INFO: first alloc address is 0x0
> BOLT-INFO: creating new program header table at address 0x400000, offset 0x400000
> BOLT-INFO: enabling relocation mode
> BOLT-INFO: forcing -jump-tables=move for instrumentation
> BOLT-INFO: enabling -align-macro-fusion=all since no profile was specified
> BOLT-INFO: enabling lite mode
> BOLT-INSTRUMENTER: Number of indirect call site descriptors: 3
> BOLT-INSTRUMENTER: Number of indirect call target descriptors: 13
> BOLT-INSTRUMENTER: Number of function descriptors: 13
> BOLT-INSTRUMENTER: Number of branch counters: 5
> BOLT-INSTRUMENTER: Number of ST leaf node counters: 15
> BOLT-INSTRUMENTER: Number of direct call counters: 0
> BOLT-INSTRUMENTER: Total number of counters: 20
> BOLT-INSTRUMENTER: Total size of counters: 160 bytes (static alloc memory)
> BOLT-INSTRUMENTER: Total size of string table emitted: 231 bytes in file
> BOLT-INSTRUMENTER: Total size of descriptors: 1500 bytes in file
> BOLT-INSTRUMENTER: Profile will be saved to file test/X86/Output/instrumentation-pie.c.tmp.fdata
> BOLT-INFO: 0 out of 16 functions in the binary (0.0%) have non-empty execution profile
> BOLT-INFO: UCE removed 0 blocks and 0 bytes of code.
> BOLT-INFO: SCTC: patched 0 tail calls (0 forward) tail calls (0 backward) from a total of 0 while removing 0 double jumps and removing 0 basic blocks totalling 0 bytes of code. CTCs total execution count is 0 and the number of times CTCs are taken is 0.
> BOLT-INFO: output linked against instrumentation runtime library, lib entry point is 0x6095c0
> BOLT-INFO: clear procedure is 0x602660
> BOLT-INFO: setting _end to 0x6006c0


And in the next command:
`bolt/test/X86/Output/instrumentation-pie.c.tmp.instrumented 1 2 3`
Output:

> Segmentation fault (core dumped)


Are you able to repro that? My host compiler is clang9. Any idea about what might be happening? I can investigate more deeply later, if you need help, let me know.

vleonen:
Hello Rafael! Thanks for checking it.
I checked how the test works with clang-9 & clang-10 on Ubuntu-20.04 and the test passed. bolt/test/X86/Output/instrumentation-pie.c.tmp.instrumented runs as expected.
Definitely, we have different input binaries produced by clang. Please share if you have any clues on what to check on my side.

rafaelauler:
(This was automatically closed because I renamed our branch from "rebased" to "main", so ignore this).

Ok, I was able to debug what is happening. In the mmap call[1] we do during initialization to share the page of our counters, the page granularity is too large for the small counters vector in that test case, so we end up mmap'ing a bunch of unrelated sections that are next in the layout (including .got section). When we do this, mmap sets all contents to zero. So, previously, we had a .got entry with the correct PLT address for "dlsym".  After instrumentation initialization code, thanks to this mmap, the address for 'dlsym" in the got table is now set to zero.  When we call "dlsym", the code jumps to zero and crashes.

I think the solution to this would be (1) to ensure that the size of our counters vector is always in multiples of the page size (0x1000). Either that, or (2) we need to add code in initialization to copy the previous contents of the page that is being mmap'ed, mmap it and then initialize it with the previously copied contents. I think (1) is simpler.

[1] https://github.com/facebookincubator/BOLT/blob/main/bolt/runtime/instr.cpp#L1465

vleonen:
Thanks for sharing your results of the issue debugging. In order to make this case more clear - could you share your binary to reproduce the issue on my side or at least "readelf -lSW" output for the file. Unfortunately, I'm not able to build the same binary with gcc or clang compilers in my environment. In my case .got section resides before .text and .bolt.instr.counters sections, so it si not corrupted.
As I can see - CountersEnd is aligned to page boundary: https://github.com/facebookincubator/BOLT/blob/main/bolt/runtime/instr.cpp#L1457
Also I see that there is a padding for the table: https://github.com/facebookincubator/BOLT/blob/main/bolt/src/RuntimeLibs/InstrumentationRuntimeLibrary.cpp#L183
So your solution (1) seems to be already in place and we need to debug it on a reproducible case.

rafaelauler:
You're right, CountersEnd is already padded. I looked again to understand what might be happening and discovered that the problem is __bolt_num_counters fetching an incorrect value in https://github.com/facebookincubator/BOLT/blob/main/bolt/runtime/instr.cpp#L1458

Is your __bolt_num_counters sane? I took a look at it here and discovered that we are not generating dynamic relocations to instruct the dynamic linker and loader to patch our GOT entries, so, for example, the got entry with the address of __bolt_num_counters has the value 0x602008, which is the correct static address of __bolt_num_counters, but after the dynamic library is loaded, it is loaded at a random location in memory (a given base address, say, 0x7ffff77f0000), and the got entry stays with the original address without the base address where the dynamic library was loaded added to it.

vleonen:
@rafaelauler 
Yes, __bolt_num_counters variable is in correct state in runtime. Could you check if there are relocations in libbolt_rt_instr.a ? In my case it looks as following:
```
$ readelf -aW ./build/lib/libbolt_rt_instr.a |grep num_counters
0000000000006716  0000005300000002 R_X86_64_PC32          0000000000000000 __bolt_num_counters - 4
0000000000006cf9  0000005300000002 R_X86_64_PC32          0000000000000000 __bolt_num_counters - 4
    83: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND __bolt_num_counters
```
Also in the runtime library, the following instruction is related to __bolt_num_counters:
```
$ objdump -dS ./build/lib/libbolt_rt_instr.a|grep 6cf7
    6cf7:       8b 05 00 00 00 00       mov    0x0(%rip),%eax        # 6cfd <__bolt_instr_setup+0x2d>
```
It is replaced with the following instruction in the instrumented test file:
```
$ objdump -dS ./build/tools/bolt/test/X86/Output/instrumentation-pie.c.tmp.instrumented |grep 409767
  409767:       8b 05 9b 88 ff ff       mov    -0x7765(%rip),%eax        # 402008 <_end+0x1988>
```
So it seems that in my environment __bolt_num_counters relocation was resolved during linking instrumentation library to output executable, but in your environment, it was not. Or maybe your libbolt_rt_instr.a doesn't contain relocations for __bolt_num_counters?

vleonen:
PR reopened #192 due to merge target branch renaming.

yota9:
@rafaelauler Hello! As I understand the problem, your instr library has a GOT table. Please take a look at the [patch](https://github.com/facebookincubator/BOLT/pull/192/commits/8256cc112249be79381972a4c9aa929d42c92140)  in reopened PR, it should fix the problem :) 
P.S. I'm not a big fan of using pragmas, but here I know only two options - to use either it, either attribute for each extern variable, since fvisibility flag does not affect them. I think it will be just easier to use pragma here to avoid this problem in the future with new variables.

rafaelauler:
That's correct, the runtime library is being generated with GOT entries and our linking mechanism doesn't handle them.
Thanks for the fix! Let me take a look at it.

PR #191, opened 2021-07-17
Fdata: Escape whitespaces in symbol names

yota9:
This patch is part of preparation for golang support. The golang symbols
might have spaces in the name (for example "type..eq.[10]interface {}").
Since fdata uses spaces as a field separator such names brakes the fdata
format, so we need to escape whitespaces and backslashes in symbol names
using the backslash character.

Vladislav Khmelevsky,


yota9:
Re-opened #182 due to branch renaming. 
Fixed @maksfb comments :)

yota9:
Gentle ping to @maksfb @rafaelauler  :)

rafaelauler:
Looks good to me! Let me start importing

bolt/test/X86/fdata-escape-chars.c
```diff
@@ -0,0 +1,69 @@
...
+  var = 3;
+}
+
+int main() {
+  static_symb_backslash_b();
+  symb_w_whitespace();
+  for (int i = 0; i < 2; ++i)
+    symb_backslash_b();
+}
+
```
rafaelauler, line 21: My compiler is putting static_symb_backslash_b after main in the output object, and it causing this test to fail because it expects BOLT to print these symbols in order. In order to make this test stable, I think we need to rewrite them in assembly and do not use host_cc to build it.  (Feel free to use it to link, though).

bolt/test/X86/fdata-escape-chars.c
```diff
@@ -0,0 +1,69 @@
...
+// REQUIRES: system-linux
+
+// RUN: %host_cc %cflags -O0 %s -o %t.exe -Wl,-q
+// RUN: llvm-objcopy --redefine-syms=%p/Inputs/fdata-escape-chars-syms.txt %t.exe
+//
+// RUN: llvm-bolt %t.exe -o %t.exe.instrumented -instrument \
+// RUN:   -instrumentation-file=%t.fdata
+// RUN: %t.exe.instrumented
+// RUN: cat %t.fdata | \
+// RUN    FileCheck --check-prefix="FDATA_CHECK" %s
```
rafaelauler, line 31: This is not being interpreted correctly by LIT. This is the output of LIT here:

  'RUN: at line 30';   cat bolt/test/X86/Output/fdata-escape-chars.c.tmp.fdata |  build/bin/llvm-bolt build/tools/bolt/test/X86/Output/fdata-escape-chars.c.tmp.exe -o tools/bolt/test/X86/Output/fdata-escape-chars.c.tmp.fdata.exe -data build/tools/bolt/test/X86/Output/fdata-escape-chars.c.tmp.fdata -print-finalized (...)
  
 It looks like it is pipeing fdata to the command in line 32? Can you check if this happens for you too?

rafaelauler (commented):
Hi @yota9! After the work on instrumentation landed, and went ahead to land this and got some issues, see the comments in the commit.

bolt/test/X86/fdata-escape-chars.c
yota9: Yes, you are right, thank you! It was made with the help of clang-format :)

bolt/test/X86/fdata-escape-chars.c
yota9: Yes, sorry, I forgot about this possibility :) I've decided to rewrite it using LLVM IR (actually just used clang -emit-llvm). AFAIU we can use host_cc to assemble and link the input, it should not optimize and reorder it at this point. I've check few compiler versions it it looks good, could you please check it? Thank you :)

rafaelauler: No problem! Sounds good!

PR #192, opened 2021-07-21
Instrumentation support for PIE & shared libs on x86_64 Linux

### LINKED ISSUE #137: Using Bolt in dynamic libary, occur Segmentation fault

vleonen:
This PR enables instrumentation support for PIE executables, shared libraries, and initial support for statically linked executables on x86_64 Linux systems.
We are using relative jumps in trampoline functions to avoid usage of absolute addresses calculated during build time on the linking stage for initialization/finalization functions and instrumentation calls. These absolute addresses are not valid for randomized addresses for PIE & shared libraries.
Also for correct operation for instrumented shared libraries, we added a set of syscalls and functions for searching a path to the current binary in /proc/self/map_files directory based on the current virtual address and symlinks in this directory. In case/proc/self/map_files directory access is restricted (e.g. in container environment), we added -instrumentation-binpath option to pass information about expected path to an instrumented binary during runtime.

Co-authored-by: Vladislav Khmelevsky vladislav.khmelevskyi@huawei.com
Co-authored-by: Elvina Yakubova elvina.yakubova@huawei.com

Fixes #137

Vasily Leonenko,


rafaelauler (changes requested):
Thanks for working on this! I've left some suggestions below, let me know what do you think.

bolt/src/BinaryContext.h
```diff
@@ -535,6 +535,9 @@ class BinaryContext {
   /// linked.
   bool IsStaticExecutable{false};
 
+  /// Set to true if the binary contains PT_INTERP header.
+  bool hasInterpHeader{false};
```
rafaelauler, line 538: Capitalize:

HasInterpHeader

bolt/src/Passes/Instrumentation.cpp
```diff
@@ -651,6 +651,12 @@ void Instrumentation::createAuxiliaryFunctions(BinaryContext &BC) {
       createSimpleFunction(
           "__bolt_fini_trampoline",
           BC.MIB->createSymbolTrampoline(FiniSym, BC.Ctx.get()));
+    } else {
+      // Create dummy return function for tranpoline to avoid issues
```
rafaelauler, line 672: tranpoline -> trampoline

bolt/src/Passes/InstrumentationSummary.h
```diff
@@ -104,9 +104,9 @@ struct InstrumentationSummary {
   /// Stores function names, to be emitted to the runtime
   std::string StringTable;
 
-  /// Our runtime indirect call instrumenter function
-  MCSymbol *IndCallHandlerFunc;
-  MCSymbol *IndTailCallHandlerFunc;
+  /// Pointer to runtime instrumentation handler
+  MCSymbol *IndCallInstrPointer;
```
rafaelauler, line 108: The "Instr" here confused me a bit because I was associating with "Instr" being "Instruction". Perhaps we can rename to IndCallCounterFuncPtr? (since it is basically just updating a counter)

bolt/src/Passes/Instrumentation.cpp
```diff
@@ -616,13 +616,24 @@ void Instrumentation::createAuxiliaryFunctions(BinaryContext &BC) {
     return Func;
   };
 
-  Summary->InitialIndCallHandlerFunction =
-      createSimpleFunction("__bolt_instr_default_ind_call_handler",
-                           BC.MIB->createInstrumentedNoopIndCallHandler());
+  BinaryFunction *IndCallHandler =
```
rafaelauler, line 625: It took me a while to understand what was happening here because we are not creating two functions but the real intent is to create two basic blocks of the same function that is coordinating entry to the code that performs instrumentation of indirect calls.

Can we rename this variable to IndCallHandlerExitBB, and add a comment explaning? Also, the methods "createInstrumentedIndCallHandler" and "createInstrumentedIndCallTrampoline" could be renamed to "createInstrumentedIndCalllHandlerExitBB" and "createInstrumentedIndCallHandlerEntryBB", respectively, so it's extra clear what we are doing here.

Regarding the comment explaining, I was thinking in something like:

// Here IndCallHandlerExitBB has the instructions to finish handling traffic
// to an indirect call. We pass it to createInstrumentedIndCallHandlerEntryBB(), which
// will decide between jumping to code in the runtime library (to account traffic)
// if the pointer is set, and then finish by jumping to the exit bb, which will return
// normal program execution.

bolt/src/Passes/InstrumentationSummary.h
```diff
@@ -104,9 +104,9 @@ struct InstrumentationSummary {
...
+  MCSymbol *IndTailCallInstrPointer;
```
rafaelauler, line 109: Same

rafaelauler (commented):
I think I finished reviewing the most important parts of this PR. Let me know what you think about the comments and thanks a lot for working on this, that was a great job.

bolt/runtime/instr.cpp
```diff
@@ -566,12 +566,98 @@ FunctionDescription::FunctionDescription(const uint8_t *FuncDesc) {
...
+        lenS++;
+      }
+      if (lenS == 0)
+        continue;
+      *strcp++;
+      while (*strcp != '\0') {
+        *strcp++;
+        lenE++;
+      }
+      assert(lenS + lenE < ADDR_SIZE, "file name too long");
```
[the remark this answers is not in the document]
ElvinaYakubova: Good idea, will do. Thank you

bolt/runtime/instr.cpp
```diff
@@ -566,12 +566,98 @@ FunctionDescription::FunctionDescription(const uint8_t *FuncDesc) {
...
+    return target_path;
+
+  curAddr = (unsigned long)__get_pc();
+  uint64_t FDdir = __open(dirPath,
+                          /*flags=*/0 /*O_RDONLY*/,
+                          /*mode=*/0666);
+  assert(static_cast<int64_t>(FDdir) > 0,
+         "failed to open /proc/self/map_files");
+
+  for (;;) {
```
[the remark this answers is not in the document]
ElvinaYakubova: Here is an example at the end of the page https://man7.org/linux/man-pages/man2/getdents.2.html and it seems like we need this loop (getdents can read to the buf only BUF_SIZE elements at once), otherwise we might lose some entries.

bolt/runtime/instr.cpp
```diff
@@ -566,12 +566,98 @@ FunctionDescription::FunctionDescription(const uint8_t *FuncDesc) {
...
+
+void *__attribute__((noinline)) __get_pc() {
+  return __builtin_extract_return_addr(__builtin_return_address(0));
+}
+
+/// Get full path to the real binary by getting current virtual address
+/// and searching for the appropriate link in address range in
+/// /proc/self/map_files
+static char *getBinaryPath() {
+  const char dirPath[] = "/proc/self/map_files/";
```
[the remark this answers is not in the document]
ElvinaYakubova: Thanks for the suggestions :)

bolt/runtime/instr.cpp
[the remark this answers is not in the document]
rafaelauler: Sorry, I missed the line

```
     if (nread == 0)
      break;
```

with the exit condition. So it reads until nread is zero. Then perhaps this outer loop could be written as:

  ```
while (nread = __getdents(FDdir, (struct dirent *)buf, BUF_SIZE) ) {
  }
```

To make the exit condition more explicit

vleonen:
@rafaelauler Thank you for reviewing it. I just have pushed a version updated according to your comments and rebased on top of current main branch. Please have a look at it.

rafaelauler:
Thanks a lot folks!

vleonen:
@rafaelauler Thank you!
I just found that readme file in root and in bolt folder become duplicated some time ago. And after my commit they become different (this PR patch 1437d30f7c8faabeaf1cf2426b559d5b498a99c3 modifies bolt/README.md). Please give me know if I need to prepare one more PR to sync them or do you plan to remove bolt/README.md to avoid further duplication and uncertainly?

rafaelauler:
Oh, good catch! Let me fix this.

PR #197, opened 2021-08-12
Fix AARCH64 ADR* relocations

yota9:
The ADRP instructions has 21 bits to store page offsets + 12 lowest bits
are zero, that give us a total of 33 bits (32 bits for address + 1 sign
bit, to address +- 4GB).

Vladislav Khmelevsky,


rafaelauler:
Good catch!

PR #198, opened 2021-08-12
[BOLT][NFC] Removed unused singletonSet

aaupov:
Remove unused code introduced a while ago (2016), with its use removed
since then.


rafaelauler:
Looks good to me!

PR #200, opened 2021-08-13
Instrumentation: Use trylock for indirect calls SimpleHashTable getter

vleonen:
This PR introduces TryLock usage for SimpleHashTable getter used in indirect calls instrumentation. TryLock usage avoids usage of sigprocmask syscalls currently used in Lock wrapper to avoid deadlocks. Currently, these syscalls cause significant overhead for instrumented binary execution.
Also, this PR adds a corresponding test case and lit/docker configuration to handle potential deadlocks in the test case.
Note: added lit configuration requires additional psutil python package, so it should be installed on build system.

Vasily Leonenko,


rafaelauler:
Hi @vleonen! Thanks for working on this! Could you squash all commits in one? The reason is because we need all test cases of a feature to be present in the same commit (following LLVM rules), so we can easily test it when dealing with this commit.

vleonen:
Hi @rafaelauler ! How about commit which modifies dockerfile? I'm not quite confident about how do you plan to merge this part to LLVM.

rafaelauler:
@vleonen you're right, it's probably easier for us in the future to leave Dockerfile modifications in a separate PR. Also, having multiple commits in a single PR is a bit harder to handle on our end, at the moment.

vleonen:
@rafaelauler 
I squashed commits with fix and test + lit config changes. Docker file changes are kept as a separate commit.

rafaelauler:
Thanks for working on that @vleonen, that's a very interesting bug. I though that https://github.com/facebookincubator/BOLT/pull/167 was supposed to fix that?

One thing I wanted to mention is that I'm getting the "Setting a timeout per test not supported. Requires the Python psutil module but it could not be found. Try installing it via pip or via your operating system's package manager." message.

I wonder if that is an issue for LLVM? I mean, the whole project builds and tests without this package being required. If we add it to BOLT, it would be the first test to require that package in the entire project, and will probably break all buildbots (see https://lab.llvm.org/buildbot/).

What we can do is perhaps add a new REQUIRES attribute to the test that checks if the host python has this package, and only run the test in that case. Well, I now looked at https://reviews.llvm.org/D64251 and discovered that we can use the tag "REQUIRES: lit-max-individual-test-time" for that. So we should probably use that too to be conformant with LLVM. Then, in our BOLT's lit.cfg, you can conditionally set the limit in the same way that the D64251 patch does for lldb/trunk/lit/lit.cfg.py (lldb's config): if not supported, you print a warning put proceed with tests. Then, your test will not run because it will fail the REQUIRES check. If the host python has psutil, it will proceed and run your test.

By the way, the added test in my system succeeds even when the modifications proposed in this PR are not applied. I tried running the test multiple times to see if it hits a deadlock, but it doesn't (I understand this might be me getting lucky, though, that's why I was asking on some details on what happened for you).

yota9:
Hello @rafaelauler ! This PR does not fix bug. The problem with #167 is that we use sigprocmask often, which results in higt CPU usage. This solution is not ideal too, the profile information won't be "stable" anymore, for example some tests that will check exec count may start failing. Maybe we should place this solution under option (although we have too much instrumentation options now IMHO) or think about other solution. 
The CPU usage for instrumentation by it self is not a big deal. But in some cases, like when we run software under container env we can check CPU usage as one of the sign that everything is working fine, and too high usage may results in container restart. That is one of the examples when #167 is not an ideal (but right) solution too :)

vleonen:
@rafaelauler  Thanks for digging into details about this PR. As @yota9  already explained - this PR is an alternative approach to fix the deadlock but with less overhead. That's why the test without other modifications is also passing due to already merged fix with signals masking.
Regarding lit config changes - I used lit-max-individual-test-time option for potential deadlock detection to avoid writing overcomplicated test which may detect deadlock by itself. I'll try to play around with "REQUIRES" flag specific to the test and share update.

rafaelauler:
Oh I see, I understand now, thanks! My thoughts are that indeed this should be guarded under a flag. For some applications, I imagine the user won't want to trade accuracy for lower cpu util. But we actually already have an option to disable some optimizations and make instrumentation slower. It is conservative-instrumentation. Under that option, you can put the old behavior that avoid deadlocks, and change its description to something more generic, such as "disable speed optimizations that sacrifice profile accuracy". With that, the user will be able to collect the profile in these two modes and take a decision on whether the extra profile accuracy is important or not.

Regarding the number that we are iterating until we give up and decide it is a deadlock: we currently poll the bus 100 times. What are the consequences of changing this number? My gut feeling is that it is giving up too early, since 100 loads happen pretty fast, but I'm not quite sure TBH.

vleonen:
Hello @rafaelauler 
I have updated MR commits according to your comments.
LIT changes split to separate commit because they are standalone and based on LLVM CL referred by you in above comment.
As for number of iterations before giving up to get a lock - the results (number of giveups) are sensitive to platform on which the test is running, on several systems available for me (different i7 4/8 gens and Xeon Cascade Lake) - NoEntry fires within ~20 times. Increasing the number of retries >=10000 increases number of NoEntry events, before that [10-1000] - number of events seems to be the same.

bolt/runtime/instr.cpp
```diff
@@ -272,10 +274,14 @@ class SimpleHashTable {
   /// avoid storing a pointer to it as part of this table (remember there is one
   /// hash for each indirect call site, so we wan't to minimize our footprint).
   MapEntry &get(uint64_t Key, BumpPtrAllocator &Alloc) {
+    if (!__bolt_instr_conservative) {
```
rafaelauler, line 277: Were you able to successfully build this? It's weird, but my compiler (clang 12) is complaining in this line with:

error: use of undeclared identifier '__bolt_instr_conservative'
    if (!__bolt_instr_conservative) {

rafaelauler (commented):
Thanks @vleonen! The changes look good, I have one comment

vleonen:
Thanks, @rafaelauler . I reproduced an issue with clang-12 and fixed it in the recent update by moving __bolt_instr_conservative declaration to the same namespace with SimpleHashTable, previously it was declared outside (where the rest of extern __bolt* variables are declared).

rafaelauler:
Thanks!

Issue #187, opened 2021-07-13
perf2bolt crashes


kmod:
This could certainly be user error due to the weird situation but I'm getting some crashes with perf2bolt.

My steps to create these files were:
- On my Ubuntu 20.04 machine, create a 16.04 docker container
- Use the 16.04 `perf` binary from an older kernel, but with the running newer kernel
- Run the profiling run
- Run perf2bolt, it crashes

Things I tried:
- I passed `-jump-tables=none` to the perf2bolt command, and it crashed at a different point
- I ran perf2bolt outside of the docker container (with a perf binary matching the kernel version) and it had the same crash
- I fed the .perf file directly to bolt with the `-p` flag, and that surprisingly worked

Here's the .perf file and .so file that it crashes on:
https://www.dropbox.com/s/39yagoome8hf70f/perf2bolt_crash.tar.gz?dl=0
And you should be able to reproduce the crash with `perf2bolt -p libpython3.8-pyston2.3.so.1.0.perf libpython3.8-pyston2.3.so.1.0.prebolt -o /dev/null`

aaupov:
Hi @kmod,

Sorry for taking so long to get to this issue. I think I was able to repro the issue. Is that the assertion that you were seeing?
```
perf2bolt: /vagrant/llvm-bolt/bolt/src/BinaryBasicBlock.cpp:335: 
void llvm::bolt::BinaryBasicBlock::removePredecessor(llvm::bolt::BinaryBasicBlock*, bool): 
Assertion `Erased && "Pred is not a predecessor of this block!"' failed.
```

kmod:
Hey sorry, it's been so long since I ran into it I don't remember the error message I got

kmod:
Ok sorry for the delay, I reproduced the issue using a more-recent version of bolt and yes I get that error

aaupov:
Should be resolved with 7d47004d228ab5d72b6e49000824a07415d4a10f

PR #201, opened 2021-08-16
[BOLT] Handle multiple pred->succ edges in removeAllSuccessors

aaupov:
For basic blocks with multiple edges to the same successor, the default
behavior of `removePredecessor` is to remove all occurrences of the
predecessor block in its predecessor list (`Multiple=true`).

Example:
A -> B (two edges)
```
A->removeAllSuccessors()
  for each successor of block A: // B twice
  // this removes both occurrences of A in B's predecessors list
  B->removePredecessor(A);
  // this invocation triggers an assert as A is no longer in B's
  // predecessor list
  B->removePredecessor(A);
```

If `removePredecessor` removes one block at a time (`Multiple=false`),
this removes blocks from A's successor list and B's predecessor list in
sync.

Resolves facebookincubator/BOLT#187.

Test plan: ninja check-bolt
Tested on inputs in facebookincubator/BOLT#187 (Pyston binary)


rafaelauler:
LGTM

aaupov:
This should be made redundant by the new normalize-cfg pass introduced recently: https://github.com/facebookincubator/BOLT/commit/d3b007dc9efeadb332f8d5f18968d7c7177591ce

PR #203, opened 2021-08-21
AArch64: Fix ADR instruction handling

yota9:
There are 2 problems found when handling ADR instruction:
1. When extracting value from the ADR instruction we need to do
it another way, then we do it for ADRP instruction.
2. When creating target expression the VariantKind should be other for
ADR instruction.

Also this PR introduces R_AARCH64_ADR_PREL_LO21,
R_AARCH64_TLSDESC_ADR_PREL21 and R_AARCH64_ADR_PREL_PG_HI21_NC
relocations support.

This patch introduces AdrPass, which will replace non-local
pointing ADR instructions with ADRP + ADD instructions sequence due to
small offset range of ADR instruction, so after BOLT magic there are no
guarantees that ADR instruction will still be in the range of
just +- 1MB from its target. Such problems are usually connected with errata 843419

The ARM mmaping symbols may have index, e.g. "$d.1", which should be
handled by bolt.

Also this patch adds ARM testing directory, which will be
used only on ARM testing servers. The common tests (non-assembler tests
which are platform-independent) might be moved from the X86 directory to
the parent one in the future, so such tests could be tested on both X86
and ARM machines.

Vladislav Khmelevsky,


ebach89:
Hello,

> 1. When extracting value from the ADRinstruction we need to do

Please, add space here ADR><instruction

ebach89 (commented):
I'd place space here:
> 1. When extracting value from the ADR><instruction

bolt/src/Relocation.cpp
```diff
@@ -147,15 +153,25 @@ uint64_t extractValueAArch64(uint64_t Type, uint64_t Contents, uint64_t PC) {
...
   case ELF::R_AARCH64_TLSDESC_ADR_PAGE21:
   case ELF::R_AARCH64_TLSIE_ADR_GOTTPREL_PAGE21:
-  case ELF::R_AARCH64_ADR_PREL_PG_HI21: {
+  case ELF::R_AARCH64_ADR_PREL_LO21:
+  case ELF::R_AARCH64_ADR_PREL_PG_HI21:
+  case ELF::R_AARCH64_ADR_PREL_PG_HI21_NC: {
     // Bits 32:12 of Symbol address goes in bits 30:29 + 23:5 of ADRP
-    // instruction
+    // and ADR instructions
+    bool IsAdr = !!(Contents >> 31 == 0);
```
ebach89, line 164: nitpicking: 
this `!!(Contents >> 31 == 0);` can be re-written as:
```
bool IsAdr = !(Contents >> 31);
``` 

But anyway, It's up to you

yota9: Hi, thank you for your comment :) But my variant is more readable for me, so I'd rather left it as-is..

ebach89: Ok, up to you

rafaelauler:
Is it possible to squash both commits into one? At the moment it is easier for us to review 1 commit per PR.

yota9:
@rafaelauler Sure, done.
P.S. I've got more patches for aarch64 with tests, so I hope we can submit this patch with aatch64 tests directory as soon as possible.. Thank you in advance!

bolt/src/Passes/AdrPass.cpp
```diff
@@ -0,0 +1,73 @@
+//===--- AdrPass.cpp ------------------------------------------------------===//
```
rafaelauler, line 1: In this pass we are fixing a linker problem, right?

From what I can tell, the user used the wrong program size model and caused the compiler to emit an ADR instruction when it shouldn't, because the target of this ADR could be out-of-range. Either that or the code was written in assembly-language directly, and the programmer used the ADR instruction incorrectly.

I'm thinking that the linker, when dealing with code that contains this restriction, could violate the range as well and fail with a relocation out-of-bounds error. Do you know if the linker will do the same relaxation to compensate for the limited ADR range? I don't think so, right? I don't think the linker is allowed to expand code inside a function in that way because expanding code may change offsets and break addresses that are already resolved by the assembler.

So my question would be, what's preventing this input from crashing the linker, instead of crashing BOLT? If it *could* crash the linker, then we shouldn't fix in BOLT, because it is only a matter of time when the linker is going to start crashing (and it is probably not crashing by luck). If it can't crash the linker, then I'm curious to understand why, and why is BOLT violating some assumption that the linker is not.

I would also rename this pass to ADRRelaxationPass, so it is more clear its intent.

rafaelauler (commented):
Thanks for working on this Vladislav! I've left some comments below, let me know what do you think.

bolt/src/Passes/AdrPass.cpp
yota9: This is very good question, which I didn't exam to be fare. I'm not 100% sure why it happens, it seems to be ld problem for real. But the problem is quite strange. For the quite big binary (redis server) with stack canary enabled I've got only 2 functions which are get canary's value with ADR instruction, not ADRP. But the thing is the ADR instruction is maintained page aligned, so ld still generated smth like this:
adr x0, #ff0000 //Page aligned
ldr x0, [x0, #page_offset]

bolt/src/Passes/AdrPass.cpp
```diff
@@ -0,0 +1,73 @@
...
+      const Relocation *Relocation = BF.getRelocationAt(Offset);
+      if (!Relocation)
+        continue;
+
+      MCPhysReg Reg;
+      BC.MIB->getADRReg(Inst, Reg);
+      std::vector<MCInst> Addr = BC.MIB->getAddress(
+          Relocation->Symbol, BC.Ctx.get(), Reg, Relocation->Addend);
+      It = BB->replaceInstruction(It, Addr.begin(), Addr.end());
+    }
```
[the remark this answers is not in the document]
yota9: I was thinking about this solution, but it will over complicate the code as it seemed to me. We will need to get symbol, addend from the current instruction, create newer ones with getAddress(), found the ADRP instruction and create new expression for it.. Maybe I'm wrong and saving offsets does not worth it, although I was thinking the ADR instruction is quite rare one and it is OK to do it like this.. I will try to figure out how to re-use information from the ADR instruction it self ,maybe it won't be to complicated to do..

bolt/src/Passes/AdrPass.cpp
yota9: @rafaelauler The errata 843419 file:///tmp/mozilla_yota90/Cortex-A53_MPCore_Software_Developers_Errata_Notice_21.pdf might cause such a behavior.. It's 3.30am, I will probably check it tomorrow..

yota9: @rafaelauler UPD I've checked the binary, in both cases the instruction address is ended with ff8 and ffc, this is exactly what errata says. The LD patches it, so BOLT is not compatible with -mfix-cortex-a53-843419 option. Of course the addresses are completely changed by the bolt and we could replace them by the ADRP + ADD change (although for this errata the add would always be 0, since LD just changed ADRP to ADR which stays page-aligned). I should probalby reformat the patch to check that the the extracted value is page-aligned and just replace ADR with ADRP (and target expression) at least for now. 
BTW this is the second time I'm thinking about such errata cases, I assume the LLVM won't fix them after BOLT? Especially if we have more compex case where ADRP could not be changed to ADR due to long distance. The other problem I thought was Intel "JCC Erratum", where extra nops could be needed to avoid it.. As I understand (however I'm not sure) currently such cases are ignored by bolt and llvm? If so are there any plans to support such problems? Thank you!

rafaelauler: Oh, now it's clear what's happening, thanks for explaining the context! Can you update the pass description to explain that we are fixing LD erratum? If I got it correctly, LD is patching an instruction to compensate for erratum 843419 and, in doing so, it is reducing the range, which is fine because LD knows the target is within range. But then BOLT comes and re-does the layout, and now we need to undo the changes done by LD and possibly recalculate which places need to be fixed according to the errata. But at the very minimum this pass is required, so the binary doesn't crash even in processors not affected by the errata.

In general, errata corrections that requires knowledge of the exact offset where an instruction is are a bit awkward to implement. The compiler usually can't do that, but the assembler can. Essentially, this requires fiddling with the assembler (LLVM's integrated assembler, the MCStreamer class), because that is one of the few components in the whole pipeline that has precise layout knowledge. When you use jcc erratum's in clang, it is essentially passing a flag to LLVM's assembler to do all the work.

In BOLT, we use LLVM's assembler as well, but whenever we need to implement an erratum fix, that means changes to it, which are outside of our control. https://reviews.llvm.org/D97982 is an example of one special assembler fragment that we introduced to aid in implementing these tricky layout fixes, but it was not accepted upstream.

In a related problem, we need precise layout to compute stubs (LongJmp.cpp). In this case, we do it outside of LLVM's assembler (in our own pass). This is quite expensive because we need to perfectly mimic the layout of the assembler and then perform adjustments. It works for fixing out-of-range control flow instructions, but it wouldn't work for erratum 843419 because we are not computing exact addresses of each instruction. Our granularity for that pass is basic-block level, so we know where each basic block is, but we don't know where each instruction is, which would be very expensive and is a task that is indeed better suited for the assembler.

If  you don't have this processor model in the fleet, disable the erratum fix in ld. If it is in the fleet and it is crashing regularly and a fix is required, try to look for how to do these NOP paddings in MCStreamer (for example, here https://github.com/facebookincubator/BOLT/blob/main/bolt/src/BinaryEmitter.cpp#L447 is an area where you might interact with the assembler (Streamer object) by requesting it to conditionally pad the instruction with NOPs if certain layout conditions are met (depending on the address, which is something the assembler can do for you).

yota9:
@rafaelauler Thank you for your comments! I've resolved them. 
Also I've found out that when using llvm-mc the mmaping symbols might be index (the arm elf documents allows it, eg $d.<smth>), I've fixed this issue here.

bolt/test/AArch64/adrpass.s
```diff
@@ -0,0 +1,42 @@
+# The second and third ADR instructions are non-local to functions
+# and must be replaced with ADRP + ADD by BOLT
+
+# REQUIRES: system-linux
+
+# RUN: %host_cc %cflags %s -o %t.exe -Wl,-q
```
rafaelauler, line 6: I guess you forgot to add llvm-mc before host_cc here?

bolt/src/Relocation.cpp
```diff
@@ -147,15 +153,25 @@ uint64_t extractValueAArch64(uint64_t Type, uint64_t Contents, uint64_t PC) {
...
     // Bits 32:12 of Symbol address goes in bits 30:29 + 23:5 of ADRP
-    // instruction
+    // and ADR instructions
+    bool IsAdr = !!(((Contents >> 31) & 0x1) == 0);
     Contents &= ~0xffffffff9f00001fUll;
     uint64_t LowBits = (Contents >> 29) & 0x3;
     uint64_t HighBits = (Contents >> 5) & 0x7ffff;
     Contents = LowBits | (HighBits << 2);
+    if (IsAdr) {
+      // ADR instruction
```
rafaelauler, line 170: I think that the "if (IsAdr)" and the comment line are a bit redundant, so I would suggest to drop this line, reformat dropping the curly braces as well

bolt/src/Passes/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 add_llvm_library(LLVMBOLTPasses
+  AdrPass.cpp
```
rafaelauler, line 2: Rename to ADRRelaxationPass (or a more descriptive name)

bolt/src/MCPlusBuilder.h
```diff
@@ -1348,6 +1350,13 @@ class MCPlusBuilder {
     return false;
   }
 
+  /// Store \p Target absolute adddress to \p RegName
+  virtual std::vector<MCInst> getAddress(const MCSymbol *Target, MCContext *Ctx,
```
rafaelauler, line 1354: I think that getAddress might be a bit confusing, as it is not immediately obvious that it is asking the backend to generate a code sequence that is going to load an address into a register. Perhaps rename it to "materializeAddress"  or "create<something>"?

bolt/src/Passes/AdrPass.cpp
```diff
@@ -0,0 +1,66 @@
...
+#include "AdrPass.h"
+#include "ParallelUtilities.h"
+
+using namespace llvm;
+
+namespace opts {
+extern cl::OptionCategory BoltCategory;
+
+static cl::opt<bool>
+    AdrPassOpt("adrpass-opt",
```
rafaelauler, line 21: Rename the flag to adr-relaxation (without the opt).

rafaelauler (commented):
Thanks, @yota9!

bolt/src/MCPlusBuilder.h
rafaelauler: Github reformatted my text. My suggestion was to use a name such as "createSomething", or "materializeAddress", whichever you prefer.

bolt/test/AArch64/adrpass.s
yota9: Sorry, removed while playing around and testing :)

yota9:
@rafaelauler Thank you, done!

bolt/src/MCPlusBuilder.h
```diff
@@ -539,15 +539,17 @@ class MCPlusBuilder {
   }
 
   virtual bool isADRP(const MCInst &Inst) const {
-    llvm_unreachable("not implemented");
```
rafaelauler: Don't remove this line, you don't need to call this for an X86 binary anymore.

bolt/src/MCPlusBuilder.h
```diff
@@ -539,15 +539,17 @@ class MCPlusBuilder {
...
     return false;
   }
 
   virtual bool isADR(const MCInst &Inst) const {
-    llvm_unreachable("not implemented");
```
rafaelauler: Same

bolt/src/MCPlusBuilder.h
```diff
@@ -1348,6 +1365,13 @@ class MCPlusBuilder {
     return false;
   }
 
+  /// Store \p Target absolute adddress to \p RegName
+  virtual std::vector<MCInst> materializeAddress(const MCSymbol *Target, MCContext *Ctx,
+                                                 MCPhysReg RegName,
+                                                 int64_t Addend = 0) const {
+    llvm_unreachable("not implemented");
```
rafaelauler, line 1372: This should still return something as some compilers may complain that the function does not return the declared return value, add a  "return {};" after llvm_unreachable. (I didn't get a warning from my compiler, though, so maybe compilers are getting smarter 😂 )

bolt/src/Passes/ADRRelaxationPass.cpp
```diff
@@ -0,0 +1,71 @@
...
+        continue;
+
+      BinaryFunction *TargetBF = BC.getFunctionForSymbol(Symbol);
+      if (TargetBF && TargetBF == &BF)
+        continue;
+
+      MCPhysReg Reg;
+      BC.MIB->getADRReg(Inst, Reg);
+      int64_t Addend = BC.MIB->getTargetAddend(Inst);
+      std::vector<MCInst> Addr = BC.MIB->materializeAddress(Symbol, BC.Ctx.get(), Reg, Addend);
```
rafaelauler, line 51: Could you run clang-format to double-check? I think this is running over the column limit.

bolt/src/Passes/ADRRelaxationPass.h
```diff
@@ -0,0 +1,40 @@
...
+
+#ifndef LLVM_TOOLS_LLVM_BOLT_PASSES_ADRRELAXATIONPASS_H
+#define LLVM_TOOLS_LLVM_BOLT_PASSES_ADRRELAXATIONPASS_H
+
+#include "BinaryPasses.h"
+
+// This pass replaces AArch64 non-local ADR instructions
+// with ADRP + ADD due to small offset range of ADR instruction
+// (+- 1MB) which could be easely overflowed after BOLT optimizations
+// Such problems are usually connected with errata 843419, the linker
```
rafaelauler, line 20: Could you provide a reference to the errata? It seems a bit out-of-context if we read just this comment.

bolt/src/Target/AArch64/AArch64MCPlusBuilder.cpp
```diff
@@ -337,41 +351,80 @@ class AArch64MCPlusBuilder : public MCPlusBuilder {
     return Expr;
   }
 
+  bool getOpNum(const MCInst &Inst,
+                unsigned &OpNum) const {
```
rafaelauler, line 355: It's not clear what operand number we are getting here. Based on the code, I guess this is selecting the target symbol? so maybe rename it to "getSymbolRefOperandNum?" Also, instead of using OpNum as an output parameter, we can just return an optional to make it simpler:

`  Optional<unsigned> getSymbolRefOperandNum(const MCInst &Inst);`

rafaelauler (commented):
Hi @yota9! Thanks, I left some comments below.

bolt/src/Target/AArch64/AArch64MCPlusBuilder.cpp
yota9: Thanks! But I think we need to stay with this argument since getTargetSymbol has one

yota9:
@rafaelauler Thank you for your comments, done

bolt/src/MCPlusBuilder.h
yota9: Well, it kind of depends on how llvm_unreachable works. Actually I prefer builtin functions (__builtin_unreachable or __builtin_trap depending on the case), which are quite common today and compiler works fine with them :)

yota9:
Hello @rafaelauler ! I hope you can import it :)

PR #204, opened 2021-08-21
Print relocations warning if failed to process

yota9:
Currently most of the warnings are printed only in debug mode. Since
relocations are very important for binary correct work I suggest to
print number of failed to process relocations to pay extra attention in
case some problems with them were met


yota9:
@rafaelauler Sure, I've removed exclamation mark

PR #207, opened 2021-09-02
Fix aarch64 TLS relocations handling

yota9:
There are few problems found when dealing with TLS relocations for
aarch64.

* RewriteInstance.cpp
** While analyzing TLS relocation we don't have to modify
SymbolAddress (which is the offset from the TLS section), so we need to
just skip verifiction
** The non-got related TLS relocations on aarch64 might be skipped like for x86
** The forse relocation must be applied for GOT relocations on
Aarch64. The symbol adress for GOT relocation might no be pointing
on GOT section (for example ADRP GOT may point to the wrong section,
since GOT table is not page-aligned), so we won't try to get section by
the symbol address.

* Relocation.cpp - Remove R_AARCH64_TLSLE_ADD_TPREL_HI12 and
R_AARCH64_TLSLE_ADD_TPREL_LO12_NC from isGOT check, since they are not
got-related relocations

* BinaryFunction.h
** Remove R_AARCH64_TLSLE_ADD_TPREL_HI12 and
R_AARCH64_TLSLE_ADD_TPREL_LO12_NC from adding to relocation list, since
this is actually an offset in TLS section and BOLT does not change it we
don't need to do something with this relocations, the value won't change
in new binary files
** Refactor the code, separating aarch64 and x86 relocations

* AArch64MCPlusBuilder.cpp
** Add forgotten LO12 relocations to switch case to getTargetExprFor

Vladislav Khmelevsky,


yota9:
Gentle ping to @rafaelauler @maksfb . I hope you will have some time to review it, since I will submit more patches soon :))

bolt/src/BinaryFunction.h
```diff
@@ -1294,51 +1294,59 @@ class BinaryFunction {
...
+      case ELF::R_AARCH64_JUMP26:
+      case ELF::R_AARCH64_TLSDESC_CALL:
+      case ELF::R_AARCH64_TLSLE_ADD_TPREL_HI12:
+      case ELF::R_AARCH64_TLSLE_ADD_TPREL_LO12_NC:
+        return;
+      default:
+        llvm_unreachable("Unexpected AArch64 relocation type in code");
+      }
+    }
+
```
rafaelauler, line 1330: I think it's cleaner to split this member function into two   "addRelocationX86" and  "addRelocationAArch64", and then call the appropriate one in addRelocation() that will have a single line in its new body:
  `return BC.isAArch64() ? addRelocationAArch64() : addRelocationX86();`
  
We already have so many functions with   if (X)  {lots of code; return}  else  {lots of code} that can be easily refactored in this way. One example is mapCodeSections that is entirely different depending on whether we are in relocation mode or not.

What do you think @maksfb? Any preference?

rafaelauler (commented):
Thanks for the test! One style comment, the rest looks good.

bolt/src/BinaryFunction.h
yota9: Done, made it in Relocation.cpp style

PR #208, opened 2021-09-06
ReorderAlgorithm.cpp: Fix iterator type

yota9:
The clang 12 doesn't want to build this place due to unrelated
types of iterator element and std vector.


rafaelauler:
Thanks!

PR #209, opened 2021-09-08
AArch64: Skip some of the relocations processing

yota9:
There are some cases, when relocations must not be processed by bolt.
This patch handles three of such cases:
*  The linker might eliminate the instruction and replace it with NOP
* The linker might perform TLS relocations relaxations, replacing the
got to direct TP + offset access.
* Due to errata 843419 the linker might create a veneer, replacing the
load/store instruction with branching.

In both cases linker leaves old relocations, that are no longer matches
the instruction emmited to binary, so we must avoid processing of these
relocations.

Vladislav Khmelevsky,


yota9:
Also we might change this approach I think. Since we know what instruction we are expect to see with relocations (eg load/store, ADRP & etc), we might check that this instruction matches the relocation and return false from analyzeRelocation if not. How do you think @maksfb @rafaelauler what approach is more correct in this case? The second one is more universal one obviously, since we won't need to actually find all the cases, when ld might change some of the instruction.. Maybe we can combine them  and also add "checkRelocation()" call after "skipRelocationProcess()" one for more flexibility and understanding what is happening in the binary file? Thank you!

bolt/src/Relocation.cpp
```diff
@@ -132,6 +132,71 @@ size_t getSizeForTypeAArch64(uint64_t Type) {
...
+  switch (Type) {
+  default:
+    break;
+  case ELF::R_AARCH64_LD64_GOT_LO12_NC:
+  case ELF::R_AARCH64_TLSIE_LD64_GOTTPREL_LO12_NC:
+  case ELF::R_AARCH64_TLSDESC_LD64_LO12: {
+    if (IsB(Contents))
+      return true;
+  }
+  }
```
rafaelauler, line 195: Can we merge this in the same switch-case statement?

yota9: Sure I may merge it, but we may have different cases with the same relocation like with R_AARCH64_TLSIE_LD64_GOTTPREL_LO12_NC . Plus it is more readable with the separated comments IMHO. Maybe we can stay with this variant? 
Also could you please answer about the other idea in the comments please?

rafaelauler: Oh I see what you mean. There is a path where R_AARCH64_TLSIE_LD64_GOTTPREL_LO12_NC fails the isMov check, and then it is checked again in the next switch-case statement. Sounds good.

rafaelauler:
> Since we know what instruction we are expect to see with relocations (eg load/store, ADRP & etc), we might check that this instruction matches the relocation and return false from analyzeRelocation if not

That is the original design of analyzeRelocation. It basically tries to process all relocations, and returns false if it finds anything weird that needs to be skipped it.  In theory, the linker can put all sorts of relocations that don't necessarily need to be consumed by BOLT, so that's normal.

My interpretation by reading this diff is that, by creating the Skip output parameter, you actually wanted to drill down on all different relocations that are left out, and mark them as failed accordingly. 

I'm fine with this approach (with the Skip output parameter) for the purposes of making it clear what we are intentionally skipping.

PR #212, opened 2021-09-22
Updating Optimizing Clang-7 tutorial (llvm-mirror)

angelica-moreira:
Minor update to the Optimizing Clang-7 tutorial, due to the sources needed to compile Clang-7 having been relocated to a new location, in the llvm-mirror branch.


rafaelauler (commented):
Hi Angelica! Thanks for the update!

Does https://github.com/llvm/llvm-project branch release/7.x work? That's the official LLVM repo now. In this case, because LLVM moved to a mono-repo, there's no need to fetch other repos separately, just:

git clone -q --depth=1 --branch=release/7.x https://github.com/llvm/llvm-project

But paths will change a bit, I think. (to llvm-project/llvm instead of just llvm/)

angelica-moreira:
> Hi Angelica! Thanks for the update!
> 
> Does https://github.com/llvm/llvm-project branch release/7.x work? That's the official LLVM repo now. In this case, because LLVM moved to a mono-repo, there's no need to fetch other repos separately, just:
> 
> git clone -q --depth=1 --branch=release/7.x https://github.com/llvm/llvm-project
> 
> But paths will change a bit, I think. (to llvm-project/llvm instead of just llvm/)

Hi Rafael, sorry for the delay, you're right I've tested and it works :). I think it is better to update it using the mono-repo then.

aaupov:
Hi Angelica! Thanks for the PR. We've incorporated it into this commit c4ed6a11eb264660f5cfe7f5165df35b63d52608 with some other changes.

angelica-moreira:
Hi Amir, happy new year :)!!! Thanks for your return.

Issue #177, opened 2021-06-23
[Question] How are`AARCH64_MOVW_UABS_G*` relocations handled?


Pierre-vh:
Hello,

While playing with BOLT on aarch64 inputs, I noticed that sometimes the old entry point is called instead of the new one when running with relocations & moving the entry point around.

The reason is that sometimes, in `_start`, compilers generate the following code:

```
   0x0000000000232018 <+24>:	movz	x0, #0x0, lsl #48
   0x000000000023201c <+28>:	movk	x0, #0x0, lsl #32
   0x0000000000232020 <+32>:	movk	x0, #0x21, lsl #16
   0x0000000000232024 <+36>:	movk	x0, #0x7a4
   0x0000000000232028 <+40>:	movz	x3, #0x0, lsl #48
   0x000000000023202c <+44>:	movk	x3, #0x0, lsl #32
   0x0000000000232030 <+48>:	movk	x3, #0x21, lsl #16
   0x0000000000232034 <+52>:	movk	x3, #0x808
   0x0000000000232038 <+56>:	movz	x4, #0x0, lsl #48
   0x000000000023203c <+60>:	movk	x4, #0x0, lsl #32
   0x0000000000232040 <+64>:	movk	x4, #0x21, lsl #16
   0x0000000000232044 <+68>:	movk	x4, #0x888
   0x0000000000232048 <+72>:	bl	0x2108d0 <__libc_start_main@plt>
   0x000000000023204c <+76>:	bl	0x2108e0 <abort@plt>
```

The first four moves are moving the address of `main` into `x0` in fragments, as aarch64 doesn't support operands that large. Those four moves have associated relocations too.
```
00000000002106c8  0000003a0000010d R_AARCH64_MOVW_UABS_G3 00000000002107b4 main + 0
00000000002106cc  0000003a0000010c R_AARCH64_MOVW_UABS_G2_NC 00000000002107b4 main + 0
00000000002106d0  0000003a0000010a R_AARCH64_MOVW_UABS_G1_NC 00000000002107b4 main + 0
00000000002106d4  0000003a00000108 R_AARCH64_MOVW_UABS_G0_NC 00000000002107b4 main + 0
```

However, BOLT doesn't seem to handle those relocations (on master) and thus the optimised executable calls the old entry point in `bolt.org.text`, leading to subtle issues.

Can this be worked around in any way, perhaps with a compiler option? 
Thanks

yota9:
Hello! 
[...]

PR #213, opened 2021-09-22
Add AARCH64_MOVW_UABS_G* relocations support

yota9:
This patch fixes issue #177


yota9:
Hello @rafaelauler ! Since without this patch the address in the register will point to the old text section. How do you think could we wipe .bolt.org.text with 0 (if we don't use use-old-text or lite mode) in order to test such cases? It also will be good in case we don't notice old text usage when we don't want it..

rafaelauler:
Hi @yota9 ! Thanks for the patch! You can use -trap-old-code for these cases. The program will crash during runtime if it is branching to old code.

yota9:
@rafaelauler Oh I didn't see this option before, thank you!

PR #219, opened 2021-10-01
Update skipRelocationProcess

yota9:
The ELF::R_AARCH64_TLSDESC_LD64_LO12 and
ELF::R_AARCH64_TLSDESC_ADR_PAGE21 relocations might also be relaxed to
mov instructions, handle these cases

Vladislav Khmelevsky,


rafaelauler:
Thanks @yota9 ! So, after this diff is applied, the tls test that is marked to fail starts passing, but because it is marked as XFAIL (expected to fail), the test actually returns a non-zero code, which is flagged as a broken test. Could you update that test to remove the XFAIL?

rafaelauler:
It looks like you already did this, I was checking an old version of your PR. Ignore this.

PR #220, opened 2021-10-03
AArch64: Add TSTBR14 and CONDB19 relocations support

yota9:
This patch adds R_AARCH64_TSTBR14 and R_AARCH64_CONDBR19 relocations
support in order to handle condition branches, cbz/cnbz and tbz/tbnz
instructions correctly


rafaelauler:
Looks good. Imported

PR #221, opened 2021-10-04
Handle relocations in constant islands

yota9:
In non-PIC binaries compiler could save absolute addresses in constant
isalnd which we should handle properly. This patch adds relocations
handling in constant islands (for AArch64).

Vladislav Khmelevsky,


rafaelauler:
Thanks for fixing this!

yota9:
@rafaelauler Thank you for reviewing this! I've updated the tests a bit. Currently I'm testing the ARM tests on X86 using cross-compiler (just a bit later I will have ARM CI and it won't be a problem anymore), it turned out the the llvm-mc doesn't know about dword directive, so I've changed it to xword. Sorry for that.

PR #222, opened 2021-10-04
Fix LongJmp pass

yota9:
This patch handles 2 problems with LongJmp pass:
1. The pass should be executed before FinalizeFunctions, since the pass
may add new entry points for the function, and the
BinaryFunction::addEntryPoint has an assert "CurrentState == State::CFG"
2. We could only relax stubs for non-PIC binaries currently, since
short/long jumps stores absolute destination jump addresses. So
currently just output error and exit for PIC binaries.

Vladislav Khmelevsky,


rafaelauler:
I think it makes sense to move the pass before the finalize pass because we are doing modifications on the functions (inserting basic blocks), and if the CFG is marked as finalized, we shouldn't be doing modifications on it.  Frame optimizations are a weird exception, but it doesn't run for AArch64, so we don't need to worry about it.

rafaelauler:
Some (internal) tests are crashing after PatchEntries got moved before finalize functions, but I didn't investigate exactly why yet. Meanwhile would you prefer to commit the previous version that does not move PatchEntries? We are working on upstreaming some internal tests, so hopefully that crash will be visible externally.

My advice would be to do not rely on LongJmp to fix PatchEntries, and to emit the longest branch sequence right off the bat.

yota9:
>> Some (internal) tests are crashing after PatchEntries got moved before finalize functions

It will be interesting to see why, I could not see a reason now..

>> My advice would be to do not rely on LongJmp to fix PatchEntries, and to emit the longest branch sequence right off the bat.

I've reverted the PatchEntries move for now and changed the tailcall instruction to shortjmp creation, I think it might be the solution, what do you think?

maksfb:
Great! Yes, it will be good to mark the instruction as a tail call for consistency.

PR #226, opened 2021-10-11
[BOLT][Instrumentation] optimization load/store X86::eflags

yavtuk:
Summary:
This commit uses reviews.llvm.org/D6629 as reference to optimize
load/store X86::EFLAGS for instrumentation snippet by using lahf/sahf
instructions instead of pushf/popf.

### DIFF
--- a/bolt/src/Passes/Instrumentation.cpp
+++ b/bolt/src/Passes/Instrumentation.cpp
@@ -177,18 +177,10 @@ Instrumentation::createInstrumentationSnippet(BinaryContext &BC, bool IsLeaf) {
   Label = BC.Ctx->createNamedTempSymbol("InstrEntry");
   Summary->Counters.emplace_back(Label);
   std::vector<MCInst> CounterInstrs;
-  CounterInstrs.resize(IsLeaf? 5 : 3);
-  uint32_t I = 0;
-  // Don't clobber application red zone (ABI dependent)
-  if (IsLeaf)
-    BC.MIB->createStackPointerIncrement(CounterInstrs[I++], 128,
-                                        /*NoFlagsClobber=*/true);
-  BC.MIB->createPushFlags(CounterInstrs[I++], 2);
-  BC.MIB->createIncMemory(CounterInstrs[I++], Label, &*BC.Ctx);
-  BC.MIB->createPopFlags(CounterInstrs[I++], 2);
-  if (IsLeaf)
-    BC.MIB->createStackPointerDecrement(CounterInstrs[I++], 128,
-                                        /*NoFlagsClobber=*/true);
+
+  BC.MIB->createInstrIncMemory(CounterInstrs, Label, &*BC.Ctx, IsLeaf,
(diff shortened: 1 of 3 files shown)


bolt/src/Passes/Instrumentation.cpp
```diff
@@ -177,15 +177,43 @@ Instrumentation::createInstrumentationSnippet(BinaryContext &BC, bool IsLeaf) {
...
+    // LAHF/SAHF are used instead of PUSHF/POPF
+    const MCPhysReg RAX = BC.MIB->getRAXRegister(8);
+    const MCPhysReg AL = BC.MIB->getRAXRegister(1);
+    // PUSHF
+    BC.MIB->createPushRegister(CounterInstrs[I++], RAX, 8);
+    BC.MIB->createClearRegisterByMovInstr(CounterInstrs[I++], RAX, 8);
+    BC.MIB->createLahf(CounterInstrs[I++]);
+    BC.MIB->createPushRegister(CounterInstrs[I++], RAX, 8);
+    BC.MIB->createClearRegisterByMovInstr(CounterInstrs[I++], RAX, 8);
+    BC.MIB->createSaveOVFlagToRegister(CounterInstrs[I++], AL);
```
rafaelauler, line 204: In general, the MCPlusBuilder interface should be target-agnostic and reflect that the user of these functions may be writing code to either X86 or Arm. But this is not always possible, so functions that are exceptions to this rule and are X86-only should be explicitly marked as "X86". So, in this case, something like "createX86SaveOVFlagToRegister" should be better.

But because this whole code sequence is pretty specific to X86 too, I think an idea that will work better in terms of design is to put the entire code sequence behind X86MCPlusBuilder through a function called "createX86FlagsPreservingIncMemory()", which returns the full sequence from Instrumentation.cpp:196 to :211, instead of adding these smaller building blocks to the MCPlusBuilder interface.  The idea of this interface is that target-specific code generation should be done in target-specific classes, instead of being done directly in the target-agnostic classes such as Instrumentation.cpp. If we move this logic to X86MCPlusBuilder, I think the design will be cleaner.

rafaelauler (changes requested):
Thanks for this patch, that's very exciting! Have you measured any performance improvements?

yavtuk:
> Thanks for this patch, that's very exciting! Have you measured any performance improvements?

Yes, I used the redis and initialization time at startup reduced by 2-3 times.

yavtuk:
> > > Thanks for this patch, that's very exciting! Have you measured any performance improvements?
> > 
> > 
> > Yes, I used the redis and initialization time at startup reduced by 2-3 times.
> 
> I am curious, what architecture is this on?

Hi Alexander,

yavtuk:
hi @rafaelauler, the new fixes is uploaded

yavtuk:
@rafaelauler give me some time to prepare another one implementation, we think it will be more readable to create two functions and move vector  resizing inside the one
Thanks

PR #228, opened 2021-10-16
Fix constant islands handling

yota9:
After the "Allocate memory for constant islands on-demand" patch there
are couple of problems found in constant islands handling:
1. When creating constant island dependency we need to check that we
already allocated IslandInfo for BF.
2. In ADRRelaxationPass we need to set constant island check under new
hasIslandsInfo condition.
3. In binaryemitter we need to replace hasConstantIsland with
hasIslandsInfo check since originally the BF might not have constant
island, but might have access to other's BF CI.


aaupov:
> Hello @maksfb @rafaelauler @aaupov ! I would like to ask about possibility of aarch64 testing using your CI. I know that the arm is not facebooks target architecture, but how do you think is it possible to organize the testing using your environment? Thank you!

Hi @yota9,

It should be possible to set up ARM testing using GitHub actions with an AArch64 VM.
I can take a look at it during the week of Oct 25. It's reasonably straightforward to do, so you're welcome to add it sooner.

PR #230, opened 2021-10-16
Handle end section labels

yota9:
We need to handle section end labels, that might be located right after
last section byte. This patch is also the part of golang support by
llvm-bolt.

Vladislav Khmelevsky,


yota9:
Hello @rafaelauler @maksfb, gentle ping on the rest of the reviews :)

maksfb:
Hi Vladislav, thanks for the ping - will get back to you soon.

PR #237, opened 2021-10-18
Disable InlineMemcpy with NoInline

yota9:
Sometimes we want completely disable inlining optimizations, disable
memcpy when NoInline option is passed

Vladislav Khmelevsky,


yota9:
@maksfb Thank you for your answer! The idea here is to protect bolt users from using incompatible options e.g. memcpy inlining and golang. Actually I didn't noticed StringOps option in pass registration, so never mind , I will just close the review :)

PR #238, opened 2021-10-19
Aarch64: Add ABS32/16 relocations support

yota9:
Add R_AARCH64_ABS32 and R_AARCH64_ABS16 relocations support

Vladislav Khmelevsky,


rafaelauler:
Unless absolutely necessary, in general we should avoid landing dead code (functions that are not used elsewhere in the codebase). Is it possible to move those functions to the PR that will make use of them?

PR #240, opened 2021-10-21
[NFC] Refactor sources to be buildable in shared mode

rafaelauler:
Summary:
Moves source files into separate components, and make explicit
component dependency on each other, so LLVM build system knows how to
build BOLT in BUILD_SHARED_LIBS=ON.

To achieve this, we create a new library to hold core IR files (most
classes beginning with Binary in their names), a new library to hold
Utils, some command line options shared across both RewriteInstance
and core IR files, a new library called Rewrite to hold most classes
concerned with running top-level functions coordinating the binary
rewriting process, and a new library called Profile to hold classes
dealing with profile reading and writing.

To remove the dependency from BinaryContext into X86-specific classes,
we do some refactoring on the BinaryContext constructor to receive a
reference to the specific backend directly from RewriteInstance. Then,
the dependency on X86 or AArch64-specific classes is transfered to the
Rewrite library. We can't have the Core library depend on targets
because targets depend on Core (which would create a cycle).

Files implementing the entry point of a tool are transferred to the
tools/ folder. All header files are transferred to the include/
folder. The src/ folder was renamed to lib/.


rafaelauler:
I changed cmake config in BOLT to only support Linux. In other platforms, we print a warning that we won't build BOLT.
I further changed configs to determine whether we will build BOLT runtime libs. This only happens in x86 host. If true, we will build the runtime and enable bolt-runtime tests. New tests that depend on the bolt_rt lib needs to be marked REQUIRES:bolt-runtime. I updated the relevant tests.

This cmake config should be more opensource friendly.

I also fixed cmake to do not crash when building llvm with a target that BOLT does not support.

Issue #206, opened 2021-08-26
BOLT-CoreDump: assert failed in long-jmp stage on aarch64 platform


Haishan312:
Hi,
thanks firstly.

using llvm-bolt(lastest main branch) to generate optimal binary for aarch64. 
```
llvm-bolt ./server_binary -o ./server_binary.bolt -data=./server_binary.fdata -reorder-blocks=cache+ -reorder-functions=hfsort -split-functions=2 -split-all-cold -split-eh -dyno-stats -v=2
```

Then I get assert failed in long-jmp pass.
```
BOLT/bolt/src/BinaryFunction.cpp:3375: llvm::MCSymbol* llvm::bolt::BinaryFunction::addEntryPoint(const llvm::bolt::BinaryBasicBlock&): Assertion `CurrentState == State::CFG && "basic block can be added as an entry only in a function with CFG"' failed.

 #9 0x00000000004af9de llvm::bolt::BinaryFunction::addEntryPoint(llvm::bolt::BinaryBasicBlock const&) /tmp/BOLT/bolt/src/BinaryFunction.cpp:3378:32
#10 0x0000000000f7c838 llvm::bolt::LongJmpPass::replaceTargetWithStub(llvm::bolt::BinaryBasicBlock&, llvm::MCInst&, unsigned long, unsigned long) /tmp/BOLT/bolt/src/Passes/LongJmp.cpp:254:19
#11 0x0000000000f7e5f6 llvm::bolt::LongJmpPass::relax(llvm::bolt::BinaryFunction&) /tmp/BOLT/bolt/src/Passes/LongJmp.cpp:579:30
#12 0x0000000000f7e964 llvm::bolt::LongJmpPass::runOnFunctions(llvm::bolt::BinaryContext&) /tmp/BOLT/bolt/src/Passes/LongJmp.cpp:618:7
```

this assert failure show we can't insert new tailcall basicblock in CFG_Finalized stage.
Therefore, I update state of Function, and this assert failure is gone.
and new basic block information:
```
.LStub1189:
    00000000:   b       SYMBOLat0x1a6bcf0 # TAILCALL
```
But in the later emit-link stage,  I encounted another assert failure.

```
resolveAArch64Relocation, LocalAddress: 0x7fdbc5b5d340 FinalAddress: 0xa1e4140 Value: 0x1a6bcf0 Type: 0x11b Addend: 0x0
[...]

PR #241, opened 2021-10-21
LongJmp: Fix alignment issues

yota9:
This patch fixes multiple problems:
* Instead of using opts::Align* we need to use per-function values
* The cold section is aligned to the maximum value that was passed to the
  emitCodeAlignment call. So we need to align DotAddress on
  tentativeLayoutRelocColdPart call and set alignment to the cold section
  in case we won't call emitCodeAlignment.
* Adjust PageSize if it is smaller then AlignFunctions value
* The rare case: Check is we can reach stub insertion point and relax
  instruction in place if we cannot.
* Small refactorings to speed-up pass execution: add breaks, add
  EstimatedTextSize under opts::UseOldText option

Vladislav Khmelevsky,

### DIFF
--- a/bolt/include/bolt/Passes/LongJmp.h
+++ b/bolt/include/bolt/Passes/LongJmp.h
@@ -31,6 +31,10 @@ namespace bolt {
 /// 64-bit range, we guarantee it can reach any code location.
 ///
 class LongJmpPass : public BinaryFunctionPass {
+  static const int NotRelaxed = 0;    // No relaxations were made
+  static const int InstrRelaxed = 1;  // Only instruction relaxations were made
+  static const int StubsInserted = 2; // Stubs were inserted
+
   /// Used to implement stub grouping (re-using a stub from one function into
   /// another)
   using StubTy = std::pair<uint64_t, BinaryBasicBlock *>;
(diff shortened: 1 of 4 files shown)


yota9:
Hello @rafaelauler ! This pass will be used with golang support. The problem with golang that we need exactly byte-to-byte binary reading from the source file and need byte-to-byte layout of the new binary before emitting it. So LLVM should not relax instructions by it's own after golang pass is finished. I've added longjmp support for x86 just recently, before that I've just relaxed all branches to long jumps in order to work with golang. So it will be activated only for golang :)
P.S. This patch also solved couple of issues in with this pass in order to make it more accurate.

yota9:
@rafaelauler Actually it is not very expensive for x86, since here we rely on instruction relaxation. So not stubs are inserted and no fixbranches are needed easier. I didn't fix shortJmp/longJmp creation for x86, since +- 2Gb are seems to be enough for us currently :)

rafaelauler:
Another thing we can do, instead of doing one larger PR, is to open a new PR with a commit on top of this one, showing how this is used. But ideally, we would have a "self-contained" PR, that is, a set of changes that we can test with a test case.

yota9:
@rafaelauler Sure, let me remove x86 support (e.g. instruction ), there are enough fixes not related to architecture support :)

Haishan312:
Can this pr fix the core in #206 ?

yota9:
@Haishan312 You should check it :) P.S. Don't use align-blocks, it is unsupported and not related to this PR.

Haishan312:
> @Haishan312 You should check it :) P.S. Don't use align-blocks, it is unsupported and not related to this PR.

But, I still get a failure information "Assertion `isInt<21>(BranchImm)' failed". and it is same to #206 
My binary is 602.gcc_s

PR #251, opened 2021-11-13
Fix ShrinkWrapping pop order

yota9:
The push and pop instructions might have wrong reorder due to this
error. Thanks rafaelauler for the provided test case.

Vladislav Khmelevsky,


yota9:
Hello @rafaelauler ! The meta-merge-fdata test is failed for some reason, could you please check it? I could not reproduce it on my env, and I don't think it is connected somehow to this patch. Thank you!

Issue #217, opened 2021-09-26
Instrumentation crashes on Rust code


lnicola:
- on `rust-analyzer` I got

```
BOLT-INFO: shared object or position-independent executable detected
BOLT-INFO: Target architecture: x86_64
BOLT-INFO: BOLT version: 0ae4e2d854c6d2435088cef682e967c59b15f628
BOLT-INFO: first alloc address is 0x0
BOLT-INFO: creating new program header table at address 0x1800000, offset 0x1800000
BOLT-WARNING: debug info will be stripped from the binary. Use -update-debug-sections to keep it.
BOLT-INFO: enabling relocation mode
BOLT-INFO: forcing -jump-tables=move for instrumentation
BOLT-INFO: enabling -align-macro-fusion=all since no profile was specified
BOLT-INFO: enabling lite mode
BOLT-WARNING: Failed to analyze 649 relocations
BOLT-WARNING: 73 collisions detected while hashing binary objects. Use -v=1 to see the list.
BOLT-INSTRUMENTER: Number of indirect call site descriptors: 135588
BOLT-INSTRUMENTER: Number of indirect call target descriptors: 36745
BOLT-INSTRUMENTER: Number of function descriptors: 36745
BOLT-INSTRUMENTER: Number of branch counters: 354311
BOLT-INSTRUMENTER: Number of ST leaf node counters: 234393
BOLT-INSTRUMENTER: Number of direct call counters: 145977
BOLT-INSTRUMENTER: Total number of counters: 734681
BOLT-INSTRUMENTER: Total size of counters: 5877448 bytes (static alloc memory)
BOLT-INSTRUMENTER: Total size of string table emitted: 5139036 bytes in file
BOLT-INSTRUMENTER: Total size of descriptors: 36112068 bytes in file
BOLT-INSTRUMENTER: Profile will be saved to file /tmp/prof.fdata
BOLT-INFO: 0 out of 39961 functions in the binary (0.0%) have non-empty execution profile
BOLT-INFO: the input contains 3793 (dynamic count : 0) opportunities for macro-fusion optimization that are going to be fixed
BOLT-INFO: UCE removed 1401 blocks and 19965 bytes of code.
BOLT-INFO: SCTC: patched 0 tail calls (0 forward) tail calls (0 backward) from a total of 0 while removing 0 double jumps and removing 0 basic blocks totalling 0 bytes of code. CTCs total execution count is 0 and the number of times CTCs are taken is 0.
[...]

PR #252, opened 2021-11-16
Disable stack protection in runtime libraries

yota9:
In some of the system stack protection is enabled by default, which will
lead in extra symbols dependencies, which we want to avoid.

Vladislav Khmelevsky,


aaupov:
Can you please elaborate the scenario in which these extra symbols create an issue (system/platform)?

lnicola:
We found this in https://github.com/facebookincubator/BOLT/issues/217#issuecomment-970089077 (compiling Rust programs on Arch Linux).

yota9:
Hello @aaupov ! On many linux distros the stack protection is enabled by default in compiler. In case it is enabled it will create some symbol dependensis e.g. __stack_chk_guard, __stack_chk_fail which are located in libc, so the binary usually will have them UND. Since the runtime libs are stand alone we don't want to create extra dependencies for them. We may add these symbols to the library, but actually I don't see the reason to support stack canary there, so I think just to disable it is good enough solution.

aaupov:
Thanks @lnicola, @yota9!
I wasn't aware of the libc dependencies, but I see now. Looks like we can't just pull __stack_chk_guard symbol into the library since it's supposed to be initialized at program load time.
I guess that's enough of a reason to disable stack-protector for runtime libraries.

aaupov:
Details of how __stack_chk_guard, __stack_chk_fail are used: https://access.redhat.com/blogs/766093/posts/3548631 (How does stackguard actually work)

yota9:
@aaupov We can do initialize it buy our selves. I mean we will have out own copy of the canary and fail function that the library will use, the other part of the code will still use LIBCs one. But I don't see the reason to do this. The canary is used for 2 things - security (not our case, since the instrumentation is not used on the production, although the hugify might) and the code debugging. The last one is a good thing, but although the instrumentation is not about performance, we want to keep the instrumentation code as fast as possible to get as much profile as possible, so in my mind it is OK not to use it here.

aaupov:
77d5dcbd399fa64f05e296ed14d50ec6d43e231c

PR #253, opened 2021-11-18
[PR]Aarch64：Add R_AARCH64_P32_ADR_PREL_LO21 relocation type

Haishan312:
Add R_AARCH64_P32_ADR_PREL_LO21 relocation type support


yota9:
Adding size for relocation is not sufficient to add support. You can see the example here: https://github.com/facebookincubator/BOLT/commit/61a2062542cd7f015cde0e0ff06abe305993943d . But also I'm not 100% sure that the BOLT will now correctly work with the ILP32 on all it's stages, this should be checked carefully..

Haishan312:
> Adding size for relocation is not sufficient to add support. You can see the example of adding relocation support here: [61a2062](https://github.com/facebookincubator/BOLT/commit/61a2062542cd7f015cde0e0ff06abe305993943d) . But also I'm not 100% sure that the BOLT will now correctly work with the ILP32 on all its stages, this should be checked carefully.. E.g. the data extraction with pointer size, maybe the RewriteInstance might have some problems, I'm not sure..

Thanks for reply. we compile spec2017 with gcc-10, and bolt report an assert stack. Then add this type, it works ok

yota9:
@Haishan312 Do you use update-debug-sections flag?

Haishan312:
> @Haishan312 Do you use update-debug-sections flag?

Yes, I add this flag in my script.

yota9:
@Haishan312 This is the real problem, not the relocation it self. I think our team is already working on this :)

Haishan312:
@yota9 That's great. Expect your patch.

PR #255, opened 2021-11-23
Fix update-debug-sections for AArch64

ElvinaYakubova:
This patch adds AArch64 relocations handling in case updating of debug sections is enabled.

Refers to https://github.com/facebookincubator/BOLT/pull/253 problem that occurs when update-debug-sections flag is used. Previously, in the updateLineTableOffsets function, only X86 relocation (number 10) was created, and further, that number was passed to the getSizeForType function, which is architecture-dependent, so it tried to find AArch64 relocation with the same number, and it was R_AARCH64_P32_ADR_PREL_LO21.

Elvina Yakubova,


maksfb:
@ElvinaYakubova, thank you for the fix. Could you please add a test case?

bolt/lib/Core/Relocation.cpp
```diff
@@ -222,6 +222,34 @@ bool skipRelocationProcessAArch64(uint64_t Type, uint64_t Contents) {
   return false;
 }
 
+
```
rafaelauler, line 225: nit: remove extra line here

bolt/lib/Core/Relocation.cpp
```diff
@@ -222,6 +222,34 @@ bool skipRelocationProcessAArch64(uint64_t Type, uint64_t Contents) {
...
+  case ELF::R_AARCH64_ABS32:
+    break;
+  case ELF::R_AARCH64_PREL32:
+    Contents -= PC;
+    break;
+  }
+  return Contents;
+}
+
+
```
rafaelauler, line 252: Remove extra line here as well

rafaelauler (commented):
Thanks for the patch Elvina! I have some suggestions, let me know what do you think.

ElvinaYakubova:
> @ElvinaYakubova, thank you for the fix. Could you please add a test case?

Sure, will do

ElvinaYakubova:
Thanks for the review @rafaelauler, fixed as you suggested. I've added tests and also noticed when trying to import line-number.test, that  "#debug line" number doesn't always show the correct value, so I'm trying to fix that as well.

bolt/lib/Core/BinarySection.cpp
```diff
@@ -148,37 +148,16 @@ void BinarySection::flushPendingRelocations(raw_pwrite_stream &OS,
...
-                        << Twine::utohexstr(SectionAddress + Reloc.Offset)
-                        << " Offset 0x"
-                        << Twine::utohexstr(SectionFileOffset + Reloc.Offset)
-                        << '\n';);
-      break;
-    }
-    }
+      TargetAddress += Resolver(Reloc.Symbol);
+
+    uint64_t Value = Relocation::getValue(Reloc.Type, TargetAddress, SectionAddress + Reloc.Offset);
```
rafaelauler, line 155: Can you run clang-format on this line? This line exceeds 80 columns.

rafaelauler (commented):
Thanks for adding the test!

bolt/lib/Core/BinarySection.cpp
ElvinaYakubova: done, thank you :)

rafaelauler:
Thanks Elvina! Sorry, I tried to import this but the linter is complaining about formatting of the changes in two other files as well (Relocation.cpp, DWARFRewriter.cpp). Could you run clang-format in the changes introduced there too?

ElvinaYakubova:
@rafaelauler Sorry for the inconvenience, I fixed it, now everything should be ok.

aaupov:
Committed: 04b81f7fe7dd3fb9dec09f6360ea4f8a1ce4082b.

PR #267, opened 2021-12-16
Dwarf: Fix high pc patching

yota9:
The DW_FORM_addr form of highPC address is written in absolute addres,
the data form is written in offset-from-low pc format.

Vladislav Khmelevsky,


yota9:
The C compilers are usually produces dwarf in with DW_FORM_data8 DW_AT_high_pc. The dwarfdump shows that the value is relative:
```
                      DW_AT_low_pc                0x00400100
                      DW_AT_high_pc               <offset-from-lowpc>218
```
 But the golang compiler produces high pc in DW_FORM_addr adress
```
                      DW_AT_low_pc                0x00457056
                      DW_AT_high_pc               0x004572e3
```

In this case we must not subtract low_pc value from high_pc.

ayermolo:
@yota9 Sorry for late reply. Can you rebase on latest. 
I redid how we handle .debug_info. It is now being re-written. So we are no longer constrained by replacement fields being the same size.

yota9:
Hello @ayermolo ! Thank you for your reply. I've rebased both the patches

bolt/lib/Rewrite/DWARFRewriter.cpp
```diff
@@ -1479,11 +1479,15 @@ void DWARFRewriter::patchLowHigh(DWARFDie DIE, DebugAddressRange Range,
     TempDebugPatcher->addLE64Patch(LowPCOffset, Range.LowPC);
   }
 
+  uint64_t HighPC = Range.HighPC;
+  if (HighPCVal->V.getForm() == dwarf::Form::DW_FORM_data8 ||
```
ayermolo, line 1483: I think high_pc can also be data2 and data1.
Would it be better to check for DW_FORM_addr since we know that is absolute and treat rest as relative?

yota9: Yes, I was thinking about this too, but forgot to do this, thank you!

ayermolo:
Would it be possible to add a test?

yota9:
Hello! For the both patches I would like to add compiled golang binary executable (actually 2, for both aarch64 and x86) where the problems were met. The thing is that the size of the binary is around 5mb each and 18mb in yaml format. The same problem would be met wet when the golang support PR will be created, since I would like to add a few more tests, since I don't think we may add dependency on go compiler. Anyway as I understand the binary size is too large to just put them in test dir, do you have any suggestions in this? Adding @maksfb @rafaelauler @aaupov to conversation :)

yota9:
Hello @ayermolo ! I've created the [test pr](https://github.com/rafaelauler/bolt-tests/pull/8/files) for both this and #268 problems. If both these reviews are good please tell me, I will create review in LLVM and then submit the patches there. Thank you!

ayermolo:
> Hello @ayermolo ! I've created the [test pr](https://github.com/rafaelauler/bolt-tests/pull/8/files) for both this and #268 problems. If both these reviews are good please tell me, I will create review in LLVM and then submit the patches there. Thank you!

@yota9 Sorry for late reply. We are in the performance reviews time ATM. :)
Does it make sense to just re-create those patches in llvm monorepo and continue review there?

yota9:
No problems @ayermolo :) Sure, let me do this tomorrow.

PR #268, opened 2021-12-16
updateDWARFObjectAddressRanges: nullify low and high pc

yota9:
In case the case the DW_AT_ranges tag already exists for the object the
low and high pc values won't be updated and will be incorrect in
after-bolt binaries.

Vladislav Khmelevsky,


yota9:
This is needed for the case if we have the input binary like this:
```
< 0><0x0000000b>  DW_TAG_compile_unit
                    DW_AT_name                  sync
                    DW_AT_language              DW_LANG_Go
                    DW_AT_stmt_list             0x00000000
                    DW_AT_low_pc                0x000a3c40
                    DW_AT_ranges                0x00000000
                ranges: 2 at .debug_ranges offset 0 (0x00000000) (32 bytes)
                        [ 0] range entry    0x00000000 0x0000129a
                        [ 1] range end      0x00000000 0x00000000
                    DW_AT_comp_dir              .
                    DW_AT_producer              Go cmd/compile go1.17.3; -shared regabi
                    <Unknown AT value 0x2905>   sync
```

The DW_AT_low_pc won't be updated and the gdb will use it's value as the range base. After the patching the range entries will be written in absolute values and the low_pc must be zeroed.
After BOLT with this patch:
```
                    DW_AT_low_pc                0x00000000
                    DW_AT_ranges                0x00000010
                ranges: 21 at .debug_ranges offset 16 (0x00000010) (336 bytes)
                        [ 0] range entry    0x00457056 0x004572e3
                        [ 1] range entry    0x00457300 0x00457344
                        [ 2] range entry    0x00457344 0x00457419
                        [ 3] range entry    0x0045741a 0x0045754c
                    .....
```

ayermolo:
Hi.

Thanks for the update. There is an internal patch that I am working on that changes how we handle .debug_info from update to "re-write". This also affects that code. Would you mind waiting until that's in and rebasing on it?

yota9:
Hello @ayermolo ! Sure, also could you please tell me what do you think about #267 patch?

ayermolo:
It looks fine. 
That whole mess with indirect will be simplified. We will be directly creating patches with forms we want.

bolt/lib/Rewrite/DWARFRewriter.cpp
```diff
@@ -672,22 +672,36 @@ void DWARFRewriter::updateDWARFObjectAddressRanges(
...
-
     std::lock_guard<std::mutex> Lock(DebugInfoPatcherMutex);
     DebugInfoPatcher.addLE32Patch(
         AttrVal->Offset, DebugRangesOffset - DebugInfoPatcher.getRangeBase(),
         AttrVal->Size);
-    if (!RangesBase)
+
+    // high_pc attribute is unused, nullify if exists
+    if (HighPCAttrInfo) {
+      if (isHighPcFormEightBytes(HighPCAttrInfo->V.getForm()))
```
yota9, line 691: @ayermolo Should I replace it with 
`TempDebugPatcher->addLE32Patch(HighPCOffset, HighPC, HighPCVal->Size);` ?

ayermolo: Good question. The size parameter is needed if we are changing the form. Say going from 8 byte form to 4 byte form. It tells the code later how many bytes to skip in the original .debug_info section when it is copying chunks of data between patches. If nothing is specified its assumed size remains unchanged. So for addLE32Patch it remains 4 bytes. 

Since we are not changing form here we need to use corresponding addLE{64,32}Patch APIs I think.

bolt/lib/Rewrite/DWARFRewriter.cpp
```diff
@@ -672,22 +672,36 @@ void DWARFRewriter::updateDWARFObjectAddressRanges(
...
+        DebugInfoPatcher.addLE64Patch(HighPCAttrInfo->Offset, 0);
+      else
+        DebugInfoPatcher.addLE32Patch(HighPCAttrInfo->Offset, 0);
+    }
+
+    if (!RangesBase) {
+      // low_pc attribute is unused, nullify if exists
+      if (LowPCAttrInfo)
+        DebugInfoPatcher.addLE64Patch(LowPCAttrInfo->Offset, 0);
```
ayermolo, line 700: What about if it's DWARF4 and debug fission or DWARF5? It will be an index in to .debug_addr section.

ayermolo:
Would it be possible to add a test for this?

bolt/lib/Rewrite/DWARFRewriter.cpp
yota9: I will need to investigate this question, I'm not extremely familiar with dwarf format to be honest :) If you have the solution for this please let me know.

ayermolo: Something like
```
if (LowForm == dwarf::DW_FORM_GNU_addr_index ||
      LowForm == dwarf::DW_FORM_addrx) {
   uint32_t AddressIndex =
        AddrWriter->getIndexFromAddress(0, *Unit->getDWOId());
        DebugInfoPatcher.addUDataPatch(LowPCOffset, AddressIndex,
                                    std::abs(int(HighPCOffset - LowPCOffset)));
}
```
Code assumes highpc/lowpc are next to each other.
Should handle DWARF4 Debug fission. Fine to not handle DWARF5 just add // TODO. I am still working on adding that support.

yota9: Hello! Sorry for late response, I had a weekend :) 
Thanks for your code, but isn't it another functionality, that is should not be it handled in separate patch? As I understand this case is not handled currently?

ayermolo: I don't quite remember what that tests tests for, but it might not just trigger this path.
Well as it stands now you will break DebugFission if this path is hit. Because you are always adding a 64bit patch.

yota9: I think I understood you, what if we'll came up with the solution like this?
```
    if (!RangesBase) {
      if (LowPCAttrInfo && HighPCAttrInfo)
        convertToRangesPatchDebugInfo(DIE, DebugRangesOffset, DebugInfoPatcher);
      else if (LowPCAttrInfo)
        DebugInfoPatcher.addLE64Patch(LowPCAttrInfo->Offset, 0);
      return;
    }

```

yota9:
@aaupov I've found one more problem, after the "Rewrite of .debug_info section" patch I've got error in dwarfdump, didn't investigate it yet: `dwarfdump ERROR:  dwarf_child:  DW_DLE_ABBREV_MISSING(394) Unable to find abbrev for DIE`

ayermolo:
@yota9 can you file a bug with repro, and assign it to me?

bolt/lib/Rewrite/DWARFRewriter.cpp
ayermolo: Not sure this will do what you want.
Why not just check for form like like I suggested, or even what you posted yesterday?
I don't think my suggestion ads a new functionality. It does same thing you are doing now: zeroing out low_pc, it just handles variable encoding used for DWARF4 Debug Fission, or DWARF5.

On side note DW_AT_high_pc and DW_AT_ranges. I am not sure that even follows the DWARF spec. When it talks about it refers either to having DW_AT_low_pc/DW_AT_high_pc or DW_AT_ranges. How is Golang uses DW_AT_high_pc, with DW_AT_ranges?

yota9:
> @yota9 can you file a bug with repro, and assign it to me?

Sure, https://github.com/facebookincubator/BOLT/issues/274 but I don't have rights to assign issues :) Thanks!

bolt/lib/Rewrite/DWARFRewriter.cpp
yota9: @ayermolo If the variant I've posted yesterday is good I will endup with this one, I've probably misunderstood you. 
Speaking of highpc check - it doesn't use this, I've made extra nullify in case it exists to handle it the same way as low pc, but probably you're right and it is not needed

ayermolo: ah. Then let's remove high_pc.
Sure, sounds good.

PR #269, opened 2021-12-17
Fix the path to the LLVM source in the CMake invocation in the Bolt README (NFC)

joker-eph:


bolt/README.md
```diff
@@ -52,7 +52,7 @@ Start with cloning LLVM and BOLT repos:
 > git clone https://github.com/facebookincubator/BOLT llvm-bolt
 > mkdir build
 > cd build
-> cmake -G Ninja ../llvm-bolt/llvm -DLLVM_TARGETS_TO_BUILD="X86;AArch64" -DCMAKE_BUILD_TYPE=Release -DLLVM_ENABLE_ASSERTIONS=ON -DLLVM_ENABLE_PROJECTS="clang;lld;bolt"
```
aaupov, line 55: This `llvm-bolt` here refers to the destination of a git clone above.

aaupov:
Thanks for spotting. We'll update the build instructions to reflect monorepo location

